PR #130, opened 2015-01-13
Update config to allow for passing of heartbeat

srt32:
On a project recently we wanted to adjust the `heartbeat` setting. This update
allows for passing in the `heartbeat` setting.


michaelklishin:
Thank you!

PR #132, opened 2015-01-15
Serialize messages for logging using JSON

drewish:
We ran into an issue where we had a lot of errors publishing messages to rabbit. We need to resend the failed messages and the content is available in the server logs, but it's in ruby's inspect format making it a huge pain to parse and resend. 

So what do you think about just logging the JSON we're sending out? It makes it much simpler to resend failed messages from logs, and you don't have to wonder about how things like dates were serialized.

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -216,9 +216,9 @@ module Hutch
       }
       properties[:message_id] ||= generate_id
 
-      logger.info("publishing message '#{message.inspect}' to #{routing_key}")
-
-      response = @exchange.publish(JSON.dump(message), {persistent: true}.
+      json = JSON.dump(message)
+      logger.info("publishing message '#{json}' to #{routing_key}")
+      response = @exchange.publish(json, {persistent: true}.
         merge(properties).
         merge(global_properties).
         merge(non_overridable_properties))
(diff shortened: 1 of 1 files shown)


michaelklishin:
This is a good idea if you ask me but still a fairly opinionated change. Any opinion from @hmarr and other @gocardless/hutch folks?

hmarr:
I'm :+1: - this is really useful, and we're jsonifying the message for transport anyway.

drewish:
Yeah agreed, it seemed like there might be some side effects but I couldn't really think of what they'd be. What drove me crazy about the ruby inspect format is you can't just drop it back into the console and have it create objects things like dates or assign values to keys ending with `?`:
```
irb(main):016:0> {has_coupon?: false}.inspect
=> "{:has_coupon?=>false}"
irb(main):017:0> {:has_coupon?=>false}
SyntaxError: (irb):17: syntax error, unexpected =>
{:has_coupon?=>false}
               ^
	from /Users/andrew/.rbenv/versions/2.1.4/bin/irb:11:in `<main>'
irb(main):018:0> {date: Time.now}.inspect
=> "{:date=>2015-01-15 14:26:44 -0800}"
irb(main):019:0> {:date=>2015-01-15 14:26:44 -0800}
SyntaxError: (irb):19: syntax error, unexpected tINTEGER, expecting '}'
{:date=>2015-01-15 14:26:44 -0800}
                     ^
	from /Users/andrew/.rbenv/versions/2.1.4/bin/irb:11:in `<main>'

```

michaelklishin:
Thank you!

PR #138, opened 2015-03-08
Support ERb in YAML config

olleolleolle:
Inspired by Rails, I found myself wanting a configuration file like this:

```yaml
uri: <%= ENV['AMQP_URL'] %>
require_paths:
  - raven
  - consumers
```

### DIFF
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -78,7 +78,7 @@ module Hutch
     end
 
     def self.load_from_file(file)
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you.

olleolleolle:
Hm. It fails on Ruby 2.1.0. Also on my machine. I was using latest stable Ruby, which was 2.2.1 at this time. 

Works on:

- 2.2.1
- 2.2.0

Does not work on:

- 2.1.0

PR #139, opened 2015-03-09
document graceful exits for hutch cli

midu:
I found this information in `worker.rb` but not in the README, which seems to be the main place to find documentation for hutch. I thought it was important enough to belong on the front page.


michaelklishin:
Thank you!

PR #140, opened 2015-03-10
Implement graceful exit

midu:
This is a follow up to #139, I realized that the documentation I added is not actually what happens.

On master, sending a SIGINT to hutch will just run `kill` on the worker pool. It's graceful in that it's killing the threads before exiting the program, but it's not really graceful in regards to what those threads are actually doing (ie, processing message).

With this PR I am implementing the following:

When sending `SIGINT`, `SIGQUIT`, or `SIGTERM` to Hutch, it will:

#### 1. Run `shutdown` on the `worker_pool`

It will set `working` to false in the pool, and enqueue an exception in the queue for each worker thread. Upon reception of the exception, the worker loop will end.

#### 2. Run `join` on all the threads

`#shutdown` kills the threads that were idling, but some workers were potentially already running so they won't get to the exception until their current job is done.

By calling `join`, we wait for `Hutch::Config[:graceful_exit_timeout]` seconds (defaults to 11 to be consistent with other timeouts default value) for the threads to finish (their next job in the queue is the exception).

#### 3. `kill` the remaining workers

After `graceful_exit_timeout` passed, we call `kill` on the worker_pool, which will kill the remaining workers that didn't finish their processing.

**note: ** I observed locally that jobs killed at step 3 are requeued. I can't find where that is defined (hutch, bunny, rabbitmq?), but I would like to add that to the README if that's indeed the expected behavior.

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -193,9 +193,12 @@ module Hutch
     end
 
     def stop
-      @channel.work_pool.shutdown # enqueues a failing job
-      @channel.work_pool.join(25)
-      @channel.work_pool.kill # kill is a job is running past the timeout
+      # Enqueue a failing job that kills the consumer loop
+      channel_work_pool.shutdown
+      # Give `timeout` seconds to jobs that are still being processed
+      channel_work_pool.join(@config[:graceful_exit_timeout])
+      # If after `timeout` they are still running, they are killed
+      channel_work_pool.kill
     end
 
     def requeue(delivery_tag)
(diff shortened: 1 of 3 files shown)


midu:
do not merge this, I am not sure this is actually working when there's a lot of messages in the queue being quickly processed. But happy to hear whether you think this is a valuable feature

michaelklishin:
@midu this is a sensible feature. Unacknowledged messages are re-queued by RabbitMQ when channel is closed (or connection is lost). I highly recommend reading RabbitMQ tutorials and Bunny documentation. This is covered with examples.

midu:
@michaelklishin your comment sounded like this is not something you want, but then you merged. Is it a mistake?

michaelklishin:
It is something we want. I just keep answering questions that are covered by the docs ;)

midu:
Oh, sorry about that. I really appreciate that you take the time to answer my questions!

PR #151, opened 2015-05-07
Channel broker

biinari:
Thread safety by not reusing bunny channels or anything based on a bunny channel across threads


biinari:
Wrong base for this PR

PR #152, opened 2015-06-04
Basic support for tracing message processing

nagas:
This adds a basic implementation for a pluggable message processing tracers.

Support for [NewRelic custom instrumentation](https://docs.newrelic.com/docs/agents/ruby-agent/features/ruby-custom-instrumentation) is included. It requires newrelic_rpm gem and I assume it's provided with Rails application.

Should be fairly easy expandable to other instrumentation sources as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,15 +105,11 @@ to learn more.
 
 ### Message Processing Tracers
 
-Tracers allow you to track message processing. You can set them using `tracer` method.
+Tracers allow you to track message processing.
 
 #### NewRelic
 ```ruby
-class FailedPaymentConsumer
-  include Hutch::Consumer
-  consume 'gc.ps.payment.failed'
-  tracer 'NewRelic'
-end
+Hutch::Config.set(:tracer, Hutch::Tracers::NewRelic)
 ```
 This will enable NewRelic custom instrumentation. Batteries included! Screenshoots available [here](https://monosnap.com/list/557020a000779174f23467e3).
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,19 @@ Hutch::Logging.logger = Rails.logger
 See this [RabbitMQ tutorial on topic exchanges](http://www.rabbitmq.com/tutorials/tutorial-five-ruby.html)
 to learn more.
 
+### Message Processing Tracers
+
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -103,6 +103,19 @@ Hutch::Logging.logger = Rails.logger
...
+### Message Processing Tracers
+
+Tracers allow you to track message processing. You can set them using `tracer` method.
+
+#### NewRelic
+```ruby
+class FailedPaymentConsumer
+  include Hutch::Consumer
+  consume 'gc.ps.payment.failed'
+  tracer 'NewRelic'
```
michaelklishin: Can we turn the argument here to be a class? I see zero value in it being a string.

lib/hutch/tracers.rb
```diff
@@ -0,0 +1,14 @@
+module Hutch
+  module Tracers
+    autoload :NullTracer, 'hutch/tracers/null_tracer'
+    autoload :NewRelic,   'hutch/tracers/newrelic'
+
+    TRACERS = {
```
michaelklishin: So how does one extend this hash? I recommend doing without all this Rails-y DSL stuff and just providing a class.

michaelklishin:
I like the idea but the implementation can be simplified. See inline commends.

README.md
nagas: Ah, just realised that this example is in fact not true as the implementation allows you to set the tracer only via a config. My mistake, sorry. I initially planned to use a consumer method for this but then thought there is no practical value in that and decided it would be more convenient to set it "globally" for all consumers via the config. What do you think?

michaelklishin: Setting it in the config makes sense.

nagas: Does your note about passing a class (instead of this Rails-y DSL) stand still in the config case? If so, we would need to "constantize" the `tracer` value from the config as we get it as a string. Or should we just explicitly warn in the README that this option can be set only using Hutch::Config#set?

michaelklishin: Still stands. Yes, we need to look the class up if it is passed as a string. I find this a lot more straightforward than having a DSL with aliases and a way to register new implementations with Hutch.

michaelklishin:
Thank you!

PR #156, opened 2015-07-03
Use HashWithIndifferentAccess instead of MultiJSON symbolize

teodor-pripoae:
Hi,

Using `symbolize: true` when loading JSON can lead to DoS since symbols are not garbage collected by ruby.

This PR replaces `MultiJSON.load(payload, symbolize_keys: true)` with `ActiveSupport::HashWithIndifferentAccess` in order to maintain backwards compatibility.

### DIFF
--- a/hutch.gemspec
+++ b/hutch.gemspec
@@ -4,6 +4,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency 'bunny', '>= 1.7.0'
   gem.add_runtime_dependency 'carrot-top', '~> 0.0.7'
   gem.add_runtime_dependency 'multi_json', '~> 1.5'
+  gem.add_runtime_dependency 'activesupport', '>= 3.0'
   gem.add_development_dependency 'rspec', '~> 3.0'
   gem.add_development_dependency 'simplecov', '~> 0.7.1'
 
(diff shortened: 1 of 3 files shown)


michaelklishin:
So [HashWithIndifferentAccess](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/hash_with_indifferent_access.rb) indeed uses string keys internally as much as possible.

My only question is: Hutch payloads are not really meant to be user-provided. Who are we trying to defend ourselves from? Sounds like from ourselves?

teodor-pripoae:
Well, I wanted to do a bit of request analyzing and storing request data (mostly headers) inside a NoSQL datastore on a different server. I'm just pushing data from each webserver into rabbitmq with a format similar to this:

```
{
  ...
  "headers": {
    "Accept": "application/json",
    "Content-Type": "application/json",
    "User-Agent:" "...",
    "SomeExtraHeaderSentByTheUser: "...."
  }
}
```

We are also developing a microservice using hutch which acts as a gateway to websocket and mobile push notification layer. Each microservice is pushing a new notification into rabbitmq and this notification service is forwarding it without caring which is the payload's format. I should not need to check that each service sending data is sending a fixed amount of nested hash keys.

PR #159, opened 2015-07-17
URI improvements

carlhoerberg:
Respect "amqps://" in the URI and use the default protocol ports (5672 and 5671) if no port is supplied in the URI


emk:
:+1: We spent quite a while trying to puzzle out why Hutch was failing to parse certain `amqp:` URLs. Thank you a lot for this patch!

michaelklishin:
Thank you, some (mock-based) tests need updating, I'll take care of that.

michaelklishin:
I'm running into obscure github issues at the moment when trying to push, will contact support and push in the next few days.

PR #162, opened 2015-08-05
Bunny logger

natesalisbury:
- Adds option to set a logger for bunny
- Moves connection params to private method


michaelklishin:
Thank you. Please rename the key to `:client_logger` as we will be introducing March Hare/JRuby support in the future.

natesalisbury:
You're welcome. The key has been renamed.

Issue #141, opened 2015-03-11
Worker pool and message distribution questions


midu:
Hi,

I am trying to figure out a few things to understand whether I am using hutch wrong, I am not understanding hutch's opinions, not understanding rabbitmq, or thinking about features that don't exist but could/should (happy to work on those).

#### 1. More than 1 consumer running at a time

I am noticing that when I run hutch, there doesn't seem to be much parallelization in terms of message consumption. I may be wrong, but I think 1 message is received and processed at a time.

I see that Bunny has a whole implementation of woker pools. and it doesn't seem like Hutch is taking advantage of it. 

From my quick experimentation, it shouldn't be much more complicated than passing `(nil, 10)` to `open_channel` on [this line](https://github.com/gocardless/hutch/blob/bbbf05f0217ae478be34cb814c17f8adeded75d0/lib/hutch/broker.rb#L107).  Of course 10 should come from Hutch::Config, but what it represents is the [`consumer_pool_size`](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L320)

#### 2. Balanced distribution from all the channels 

This one I am having a hard time understanding, but here is what I see:

I have two queues, A and B with many jobs in each of them. When Hutch is running, it FIRST consumes messages from queue A, before starting consuming from queue B.

My understanding is that queue A and B should be publishing messages to its subscribed consumers at a relatively balanced rate. Not one after the other.

I don't know taht much about rabbitmq/bunny/hutch so I am having a hard time figuring out where this is coming form or if I am even expecting the right thing.

The only thing I could find about this was:

> When doing so, it is important to understand that, in AMQP 0-9-1, messages are load balanced between consumers and not between queues.

This was in the [direct exchange](http://www.rabbitmq.com/tutorials/amqp-concepts.html) section. I think it applies here too since we passed the exchange step at this point...

I would really appreciate any comment about this. Happy to work on whatever the repo maintainers think could make its way into the codebase, but I'd need a little direction.

Thanks!

michaelklishin:
Bunny uses consumer thread pool of size 1 by default. Why? Because with concurrent processing you get race conditions and most Ruby developers are not prepared for that. Hutch can expose a setting to configure this.

The 2nd thing makes no sense. Hutch uses a single channel without any prefetch limit by default (configurable with `:channel_prefetch`) and there is no guarantee of ordering of deliveries for different consumers.

[...]

PR #163, opened 2015-08-11
Add worker_pool_size option for multithreaded consumers

dkastner:
Add option to set bunny's worker pool size to allow for multithreaded consumers. Will probably want to add standard disclaimers about concurrency and databases, etc.

Related to https://github.com/gocardless/hutch/issues/141

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -74,8 +74,8 @@ module Hutch
     end
 
     def open_channel!
-      logger.info 'opening rabbitmq channel'
-      @channel = connection.create_channel.tap do |ch|
+      logger.info "opening rabbitmq channel with #{channel_work_pool_size} workers"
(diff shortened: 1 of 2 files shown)


michaelklishin:
This is not a worker count but a consumer work (thread) pool. I'd rename the option to `consumer_concurrency`.

dkastner:
Ah ok, how about `consumer_pool_size`, the [name that the Bunny gem uses](http://reference.rubybunny.info/Bunny/Session.html#create_channel-instance_method)?

michaelklishin:
@dkastner that works, too. I forgot for a moment that Hutch uses the term "consumer" as well. Not all similar projects that build on top of Bunny do :)

michaelklishin:
Thank you!

dkastner:
Thanks for the quick response!

PR #164, opened 2015-08-12
Add MarchHare support

teodor-pripoae:
Hi,

This PR ads [march_hare](https://github.com/ruby-amqp/march_hare) support. I don't consider it complete, since I didn't test all the cases. The test suite passes, and I also tested it using my application that was already using hutch. 

However, I didn't test all the options passed to `MarchHare.connect` since I don't have an app to test it. Also I'm not sure if I'm properly using the signals inside the main loop, but it seems to work. 

The only part which I'm not sure how to do is handling `ConnectionError`. When rabbitmq is stopped, or there is a network failure, it seems that MarchHare is calling the shutdown handler, not trying to reconnect.

I'm sorry if I'm missing something, this is my first contact with MarchHare internals.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,9 @@ rvm:
   - "2.2"
   - "2.1"
   - "2.0"
+  - "jruby-9.0.0.0"
 services:
   - rabbitmq
+matrix:
+  allow_failures:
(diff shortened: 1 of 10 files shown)


michaelklishin:
Thank you!

PR #166, opened 2015-09-01
Added two missing SSL config options to pass on to bunny

kennon:
verify_peer and tls_ca_certificates

These are probably obscure enough to not need cli option support, but are very important when using self-signed certificates etc.

Verify_peer defaults to true (the same as the bunny default) and tls_ca_certificates is passed on if you actually set it.


michaelklishin:
Thank you. Please update `ChangeLog.md` and credit yourself!

Issue #165, opened 2015-08-26
Feature Request/Question: Sending manual ack/nacks with Hutch?


tvaillancourt:
Hey there,

Thanks a lot for Hutch! I am really enjoying it. I am looking to do something that may or may not be possible today, if not, I am glad to help make it happen.

I see that broker.rb has a "nack" and "ack" method, however they seem to only be used by Hutch internally in "worker.handle_message" (maybe others). I "think" Hutch will only send a "nack" if it has a problem handling the message and not if it fails to complete the "work" the message entails.

Would there be a way for me to "nack" or "ack" from within Worker's job itself, vs having Hutch do it automatically? Specifically, would it be possible to implement this logic:

A) A worker who executes without exceptions sends an "ack" after successful execution.
B) A "nack" is sent if the worker code hits any exception.

My goal here would be that if my code hits an exception, a job is not lost, but later retried. Any advice appreciated!

Cheers,

Tim

michaelklishin:
@tvaillancourt Hutch currently acks successfully processed messages automatically. So what are you looking to do, ack manually or let Hutch do it?

Note that re-queueing on exception can (and frequently will) lead to very long or endless redelivery loops.

tvaillancourt:
Thanks Michael!

I think what I'd specifically like to do is have Hutch send a "nack" if my Hutch worker encountered a Ruby Exception while working on the message. It would be cool if this was automatic or could be defined as some sort of Exception handler, but manual would be fine also.

Ultimately, my goal is to keep messages if my worker code is faulty or I cannot process the message somehow. In my scenario, a redelivery loop is better than a lost message, essentially.

Cheers,

Tim

michaelklishin:
After reviewing #168, I think there should be 3 acknowledgement modes:

 * `automatic` (as in Bunny)
 * `manual` (as in Bunny)
 * `framework` (current default, Hutch acks unless there was an exception)

The 3rd option may include what @teodor-pripoae suggested in #169.

Thoughts?

teodor-pripoae:
I think we can handle user defined callbacks on consumer level, similar to activerecord. I'm not sure the if the use case should be global or consumer based.

First option is the most clean one but not extensible, the 2nd and 3rd are more verbose.

```ruby
class MyConsumer
  include Hutch::Consumer

  on_failure :requeue
  # or  
  on_failure :failure_callback
  # or
  on_failure Proc.new{ requeue! }

  def failure_callback
    requeue!
  end
end
```

[...]

PR #168, opened 2015-09-01
`#ack` following `#requeue` causes exit

dkastner:
I have a hutch worker similar to the following:

``` ruby
class MyConsumer
  queue 'foo'

  def process(message)
    begin
      # ...
    rescue => e
      if e.is_a?(EphemeralError)
        requeue!
      else
        raise e
      end
    end
  end
end
```

This causes the process to shut down gracefully. Apparently, the `requeue!` sends off a `@channel.reject(delivery_tag, true)` and then exits the `Consumer` process method. Within the `Worker#handle_message` method, it then calls `broker.ack(delivery_info.delivery_tag)`. After this point the channel is closed, which seems to indicate that rabbitmq drops the connection when an ack follows a reject.

I created a test to show that ack is called after a reject.

We need some way to prevent this. A couple of ideas:

1) Add a property to `Hutch::Message` that indicates if there was already some ack/nack/reject action taken on the message within the process method. If so, do not attempt ack/nack from `Worker#handle_message`.

2) Remove the nack from the rescue block of `Worker#handle_message`. Instead, rely on a default `ErrorHandler` to do the nack and allow users to specify their own error handler that could examine the error and decide wither to nack/reject/requeue. This might have to cause a breaking change for anyone who manually sets `Hutch::Config.set :error_handlers, [..]`

### DIFF
--- a/spec/hutch/worker_spec.rb
+++ b/spec/hutch/worker_spec.rb
@@ -62,6 +62,24 @@ describe Hutch::Worker do
       worker.handle_message(consumer, delivery_info, properties, payload)
     end
 
+    context 'when the consumer requeues a message' do
+      class Rejecter
+        include Hutch::Consumer
+
+        def process(message)
+          requeue!
+        end
+      end
+
+      it 'requeues the message', :focus do
+        expect(broker).to_not receive(:ack)
+        expect(broker).to_not receive(:nack)
+        expect(broker).to receive(:requeue)
+
+        worker.handle_message(Rejecter, delivery_info, properties, payload)
+      end
+    end
+
     context 'when the consumer raises an exception' do
       before { allow(consumer_instance).to receive(:process).and_raise('a consumer error') }
 


michaelklishin:
You cannot double ack/reject a delivery. Have you checked RabbitMQ logs?

teodor-pripoae:
@michaelklishin I think he's saying that there is no method for requeuing a message. He is calling the requeue message and then worker acks the message. 

```ruby
begin
  message = Message.new(delivery_info, properties, payload)
  consumer_instance = consumer.new.tap { |c| c.broker, c.delivery_info = @broker, delivery_info }
  with_tracing(consumer_instance).handle(message)
  broker.ack(delivery_info.delivery_tag)
rescue StandardError => ex
  broker.nack(delivery_info.delivery_tag)
  handle_error(properties.message_id, payload, consumer, ex)
end
```

michaelklishin:
This is directly relevant to #165 then.

@teodor-pripoae submitted a proposed solution in #169. If we merged it, #165 would be hairier. I am not sure this is really an improvement and how relevant it would be if we introduce user-driven acknowledgements.

I consider this a duplicate of #165. What the user really wants here is the ability to ack/nack deliveries explicitly.

PR #169, opened 2015-09-02
Do not ack/nack if consumer rejects/requeues message

teodor-pripoae:
Fixes #168


michaelklishin:
Thank you for submitting this.

This is directly relevant for #165. If we introduce this, #165 would be hairier. I am not sure this is really an improvement and how relevant it would be if we introduce user-driven acknowledgements.

I'll close this in favour of #165.

PR #170, opened 2015-09-02
Allow configuration of exchange options

dkastner:


lib/hutch/config.rb
```diff
@@ -13,6 +13,7 @@ def self.initialize(params={})
         mq_host: 'localhost',
         mq_port: 5672,
         mq_exchange: 'hutch',  # TODO: should this be required?
+        mq_exchange_options: {},
```
rpond-pa: Some unsolicited advice, I would make this default to `{ durable: true}` instead of blank.

That way, anyone who upgrades won't get any big surprises, as in, their queues aren't durable anymore.

lib/hutch/broker.rb
```diff
@@ -52,10 +52,12 @@ def set_up_amqp_connection
       open_channel!
 
       exchange_name = @config[:mq_exchange]
+      default_options = { durable: true }
```
michaelklishin: This can be inlined.

lib/hutch/broker.rb
```diff
@@ -52,10 +52,12 @@ def set_up_amqp_connection
...
+      exchange_options = default_options.merge @config[:mq_exchange_options]
```
michaelklishin: Please use parentheses around method arguments.

michaelklishin: I think this entire line can be inlined, too.

lib/hutch/config.rb
michaelklishin: Agreed.

teodor-pripoae: No need for this since you are already doing the hash merge thing in broker.rb above.

michaelklishin: It makes more sense to me to move new default into the config.

dkastner: I thought about that but then there might be people who directly set `Hutch::Config.set :mq_exchange_options, custom: :thing` and then mistakenly lose the default durable option

dkastner: Also it's a pain to have to remember to re-set what is a default option

teodor-pripoae: Since this is a new functionality I think we can put a warning comment near this option. If somebody looks to find which is the required configuration key for this, will see that default is durable and also will see the comment.

michaelklishin:
Thank you. Sorry that has gone under my radar.

PR #171, opened 2015-09-03
Implement pluggable serializers

dima-exe:
Hutch uses hardcoded JSON encoder/decoder for serialization. This patch adds pluggable serializers,  Protobuf/Thrift or something else may be used with Hutch.


michaelklishin:
Can we do without this Rails-y DSL for managing serialisers? It's just a damn pair of functions, so lets assign them in the config and not waste CPU resources on every message published to look it up via a fancy DSL.

dima-exe:
Done

lib/hutch/broker.rb
```diff
@@ -358,5 +369,9 @@ def generate_id
     def global_properties
       Hutch.global_properties.respond_to?(:call) ? Hutch.global_properties.call : Hutch.global_properties
     end
+
+    def serializer(content_type)
```
teodor-pripoae: Do you still need this method ?

michaelklishin:
@dima-exe thank you. Please update [ChangeLog.md](https://github.com/gocardless/hutch/blob/master/CHANGELOG.md) and credit yourself!

dkastner:
I just have to say it's awesome stuff like this that makes hutch such a great tool to work with. Thank you!

PR #177, opened 2015-10-18
Customizable error response handling (custom nacks)

dkastner:
Code related to https://github.com/gocardless/hutch/issues/165

By default a nack is always sent on an exception.

Allows configuring custom handlers similar to the way logging is configured:

``` ruby
class Requeue
  def handle(delivery_info, properties, broker, e)
    if e.is_a?(Net::HTTPRetriableError) && !delivery_info.redelivered
      broker.requeue delivery_info.delivery_tag
      true
    else
      # Pass and let default nack handler nack the message
      false
    end
  end
end

Hutch::Config.set :error_acknowledgements, [
  Requeue.new
]
```

### DIFF
--- a/lib/hutch/acknowledgements/nack_on_all_failures.rb
+++ b/lib/hutch/acknowledgements/nack_on_all_failures.rb
@@ -1,0 +1,18 @@
+require 'hutch/logging'
+
+module Hutch
+  module Acknowledgements
+    class NackOnAllFailures
+      include Logging
+
+      def handle(delivery_info, properties, broker, ex)
+        prefix = "message(#{properties.message_id || '-'}): "
+        logger.debug "#{prefix} nacking message"
+
+        broker.nack delivery_info.delivery_tag
+
+        true
+      end
+    end
+  end
+end
--- a/lib/hutch/worker.rb
+++ b/lib/hutch/worker.rb
@@ -1,6 +1,7 @@
 require 'hutch/message'
 require 'hutch/logging'
 require 'hutch/broker'
+require 'hutch/acknowledgements/nack_on_all_failures'
 require 'carrot-top'
 
 module Hutch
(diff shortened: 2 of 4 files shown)


lib/hutch/worker.rb
```diff
@@ -129,11 +130,23 @@ def handle_error(message_id, payload, consumer, ex)
       end
     end
 
+    def acknowledge_error(delivery_info, properties, broker, ex)
+      acks = error_acknowledgements +
+        [Hutch::Acknowledgements::NackOnAllFailures.new]
```
michaelklishin: Why not simply have a non-empty default for `error_acknowledgements `?

michaelklishin:
The overall idea is sound. However, I find the way you fall back to the default a bit odd. Once that is addressed, I think we can move forward with this. Thank you!

lib/hutch/acknowledgements/nack_on_all_failures.rb
```diff
@@ -0,0 +1,18 @@
+require 'hutch/logging'
+
+module Hutch
+  module Acknowledgements
+    class NackOnAllFailures
+      include Logging
+
+      def handle(delivery_info, properties, broker, ex)
```
michaelklishin: Since the interface here is a single function, it makes more sense to use `call` so that functions (blocks, procs, etc) can be used as well.

michaelklishin: Scratch the above. It would lead to inconsistency with error handlers, which use `handle` and it is way too late to change that.

lib/hutch/worker.rb
michaelklishin: Answering my own question: because `:error_acknowldgements` is not a list, it's a chain of responsibility. Why so? Because double-acking would result in a channel exception and make underlying channel useless.

michaelklishin:
@dkastner merged with some minor tweaks. Thank you!

dkastner:
Thanks for taking a look!

PR #180, opened 2015-11-27
bunny accepts logger parameter

huanzhang:
According to bunny's connection document[0], Bunny#new accepts :logger parameter rather than :client_logger.

[0] http://rubybunny.info/articles/connecting.html


michaelklishin:
Good catch!

PR #181, opened 2015-12-25
Add Ruby 2.3 to Travis build

greysteil:
That.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,17 @@
 language: ruby
+
 rvm:
+  - "2.3.0"
   - "2.2"
   - "2.1"
   - "2.0"
   - "jruby-9.0.0.0"
+
+sudo: false
+
 services:
   - rabbitmq
+
 matrix:
   allow_failures:
     - rvm: jruby-9.0.0.0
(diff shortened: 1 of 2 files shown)


greysteil:
Didn't know that, but makes perfect sense. I've added them back!

.travis.yml
```diff
@@ -1,11 +1,17 @@
 language: ruby
+
 rvm:
+  - "2.3.0"
   - "2.2"
   - "2.1"
   - "2.0"
   - "jruby-9.0.0.0"
+
+sudo: false
```
greysteil: This moves us over to Travis's newer container-based infrastructure.

circle.yml
```diff
@@ -1,3 +0,0 @@
-machine:
```
greysteil: Don't think we have any need for Circle in addition to Travis.

.travis.yml
```diff
@@ -1,11 +1,17 @@
 language: ruby
+
 rvm:
+  - "2.3.0"
```
greysteil: Travis doesn't have a `2.3` binary yet, so we need to specify the full version here.

greysteil:
@michaelklishin - are you :+1: ?

michaelklishin:
@greysteil yes, it looks good. I am not sure why we also have Circle CI, by the way.

michaelklishin:
Thank you!

PR #185, opened 2016-01-21
Print load error exception in hutch executable

jesjos:
I encountered problems in my consumer (failing to `bundle exec`) and found them hard to debug without output.


michaelklishin:
Agreed, good catch!

Issue #191, opened 2016-02-12
Publish without sharing a channel between threads


bliof:
Isn't there a problem with `Hutch.publish`? It should [not share the channel between threads](http://rubybunny.info/articles/concurrency.html). (e.g.  changing the `consumer_pool_size`)

How about adding a channel and exchange, that will be used for publishing, in `Thread.current`?

Disclaimer: I've never used the gem, just skipping through the code. I may be missing something.

michaelklishin:
You are right, `Hutch.publish` does not assume concurrent publishing and it's a no-no with Bunny and pretty much every other client. Feel free to look into extending Hutch to fit into those cases but `Hutch.publish` is primarily meant to be used from Web apps which in Ruby are quite often single threaded for most intents and purposes.

teodor-pripoae:
@michaelklishin I'm using hutch with Puma webserver and also inside sidekiq which are both multithreaded. I'm pretty sure it is a common use case.

I've been running Hutch.publish inside sidekiq for a long time and I didn't have problems. What type of problem is expected due to sharing the same connection ? As far as I know it's not possible to concurrently send Hutch.publish when running inside MRI due to the GIL.

bliof:
@teodor-pripoae The GVL (or GIL) gives guarantees that only MRI's native C functions are atomic, not the Ruby code. Also note that you should not share the channels. You have one connection with multiple channels. I am not sure of what problems may occur. You may mix the [frames of the messages](http://rubybunny.info/articles/concurrency.html#sharing_channels_between_threads) but I don't known what will happen then (but it should be bad - maybe an error because of invalid message or maybe the messages could mix their metadata/payload (there is no mention of checksum/hash at first glance in the [spec](https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf)) :))

@michaelklishin I'll try to write a patch next week.

michaelklishin:
@teodor-pripoae unfortunately the fact that you haven't run into a concurrency hazard with Hutch doesn't mean they are not possible.

The GIL prevents parallelism but not concurrency. It doesn't guarantee much w.r.t. correct frame ordering on the wire. Every `basic.publish` frame is followed by at least one, and most often two, frames:

```
[basic.publish ch=1][content header ch=1][payload chunk ch=1]*
```

with concurrent publishing on a shared channel it can become

```
[...]

PR #194, opened 2016-02-19
Publish from separate channels per thread

bliof:
Add Hutch::Broker#publish_channel and Hutch::Broker#publish_exchange
that setup a new channel and exchange per thread when used.
Change Hutch::Broker#publish to use #publish_exchange and #publish_channel

resolve gocardless/hutch#191

I am not sure what to do with the part that shares the first channel with current thread in `Hutch::Broker#set_up_amqp_connection`. If it is removed, there will be a clear separation between channels used for publishing and the channel that is used for subscribing.

@michaelklishin

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -7,7 +7,7 @@ module Hutch
   class Broker
     include Logging
 
-    attr_accessor :connection, :channel, :exchange, :api_client
+    attr_accessor :connection, :channel, :api_client
 
     def initialize(config = nil)
       @config = config || Hutch::Config
@@ -42,7 +42,7 @@ module Hutch
       clear_thread_store
       @channel.close    if @channel
       @connection.close if @connection
-      @channel, @connection, @exchange, @api_client = nil, nil, nil, nil
+      @channel, @connection, @api_client = nil, nil, nil, nil
     end
 
     # Connect to RabbitMQ via AMQP. This sets up the main connection and
@@ -52,11 +52,9 @@ module Hutch
       open_connection!
       open_channel!
 
-      @exchange = set_up_exchange(@channel)
-
-      if !thread_store[:publish_channel] and !thread_store[:publish_exchange]
+      if !thread_store[:publish_channel]
         thread_store[:publish_channel] ||= @channel
-        thread_store[:publish_exchange] ||= @exchange
+        exchange # declare the exchange
       end
     end
 
@@ -64,8 +62,12 @@ module Hutch
       thread_store[:publish_channel] ||= set_up_publish_channel
     end
 
-    def publish_exchange
-      thread_store[:publish_exchange] ||= set_up_exchange(publish_channel)
+    def exchange
+      thread_store[:exchange] ||= set_up_exchange(publish_channel)
+    end
+
+    def exchange=(ex)
+      thread_store[:exchange] = ex
     end
 
     def open_connection!
@@ -147,7 +149,7 @@ module Hutch
         queue_bindings.each do |dest, keys|
           keys.reject { |key| routing_keys.include?(key) }.each do |key|
             logger.debug "removing redundant binding #{queue.name} <--> #{key}"
-            queue.unbind(@exchange, routing_key: key)
+            queue.unbind(exchange, routing_key: key)
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/hutch/broker.rb
```diff
@@ -219,12 +220,12 @@ def publish(routing_key, message, properties = {}, options = {})
         "publishing #{spec} to #{routing_key}"
       }
 
-      response = @exchange.publish(payload, {persistent: true}.
+      response = publish_exchange.publish(payload, {persistent: true}.
```
michaelklishin: This name makes little sense. Exchanges are not used for anything besides publishing.

michaelklishin:
Using thread-local storage is potentially breaking and not necessarily what I had in mind for #191. I think the user should have control over whether it is done, e.g. there can be multiple `Hutch::Broker` implementations or a setting.

lib/hutch/broker.rb
```diff
@@ -51,15 +52,22 @@ def set_up_amqp_connection
       open_connection!
       open_channel!
 
-      exchange_name = @config[:mq_exchange]
-      exchange_options = { durable: true }.merge @config[:mq_exchange_options]
-      logger.info "using topic exchange '#{exchange_name}'"
+      @exchange = set_up_exchange(@channel)
```
michaelklishin: Either this has to go, or the per-thread `Bunny::Exchange` instance does.

michaelklishin:
It's not necessary to store `Bunny::Exchange` instances in thread-local storage because exchanges themselves simply delegate publishing to `Bunny::Channel`. If channels are not shared, that's all you care for.

michaelklishin:
One thing this PR may or may not have considered is how automatic recovery will work with this: you are creating a thread-local cache, after all.

But looks like with Bunny that is going to work fairly transparently:

 * `Bunny::Session` [recovers "in place"](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L664) (there are no wrapping "autorecovering" connections and "dumb" connections as in Java and .NET clients)
 * `Bunny::Channel`s are reopened [in place](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/session.rb#L698) as well

so thread-local references to channels should still reference the new re-opened channels after recovery.

michaelklishin:
I'm curious what @gocardless/gocardless-core-team thinks of this change. It's fairly drastic if you ask me, and doesn't give the user much choice.

michaelklishin:
Does the community have an opinion on this? I feel this change is too opinionated and we should instead provide a way for Hutch to use a connection and/or channel that the user set up.

KushalP:
Chiming in as one of @gocardless/gocardless-core-team.

> I feel this change is too opinionated and we should instead provide a way for Hutch to use a connection and/or channel that the user set up

I agree with this thinking, but I'd like to spend a bit of time testing this change in a few different scenarios. Do we have some examples of common use cases?

michaelklishin:
@KushalP I think using Hutch in an app running on Puma is one good example. Any multi-threaded app that needs to publish via Hutch, really.

KushalP:
I've created a small repo to test some of our assumptions around this. You can find it here: https://github.com/KushalP/hutch_thread_channels

For the time being it sets up `puma` to use [multiple workers and threads](https://github.com/KushalP/hutch_thread_channels/blob/master/puma.rb#L1-L2). I've been running `wrk` against the server to get it to publish a lot of messages, but haven't been able to reproduce the issue in question yet.

Is the issue specifically about `puma` or is it to do with manually creating threads in Ruby?

michaelklishin:
It is with Hutch using a shared channel fir what it does.
Channels must not be shared between threads.

bliof:
I'll try to split this pull request into 2-3.
* this one that stores channels in a thread-local storage
* one that allows setting custom channels/exchanges with a proc

```
Hutch.exchange = proc { Thread.current[:my_exchange] ||= Hutch.broker.connection.create_channel.topic('my-test') }
Hutch.exchange # => will call the proc every time
```
* (maybe) one with specs

michaelklishin:
Sounds promising. Thank you.

michaelklishin:
Closing in favour of #199, which makes it possible to subclass `Hutch::Broker` and implement this or another concurrency hazard safety strategy.

PR #197, opened 2016-02-29
Log message using level DEBUG instead of INFO

jesjos:
This may be a matter of personal preference, but I think defaulting to DEBUG makes much more sense


michaelklishin:
I tend to think this is a good idea. @KushalP WDYT?

KushalP:
Yup. Looks good :+1:

michaelklishin:
@jesjos thank you. Please submit a PR that updates change log and credit yourself ;)

PR #199, opened 2016-03-02
Versatile broker

### LINKED ISSUE #191: Publish without sharing a channel between threads

bliof:
If the Hutch::Broker is changed like this, one could easily modify it in a way that will resolve #191 e.g.

``` ruby
# Pseudo code I have not tested it :)

class CustomBroker < Hutch::Broker
  def exchange
    return Thread.current[:exchange] if Thread.current[:exchange] 
    declare_exchange!
  end

  def declare_exchange!(ch = open_channel)
    Thread.current[:exchange] = declare_exchange(ch)
  end

  def disconnect
    Thread.list.each { |t| t[:exchange] = nil if t[:exchange] }
    super
  end
end

def Hutch.connect(options = {}, config = Hutch::Config)
  unless connected?
    @broker = CustomBroker.new(config)
    @broker.connect(options)
  end
end
```

follow-up of #194 
see issue #191

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -50,38 +50,53 @@ module Hutch
     def set_up_amqp_connection
       open_connection!
       open_channel!
-
-      exchange_name = @config[:mq_exchange]
-      exchange_options = { durable: true }.merge @config[:mq_exchange_options]
-      logger.info "using topic exchange '#{exchange_name}'"
-
-      with_bunny_precondition_handler('exchange') do
-        @exchange = channel.topic(exchange_name, exchange_options)
-      end
+      declare_exchange!
     end
 
-    def open_connection!
+    def open_connection
       logger.info "connecting to rabbitmq (#{sanitized_uri})"
 
-      @connection = Hutch::Adapter.new(connection_params)
+      connection = Hutch::Adapter.new(connection_params)
 
       with_bunny_connection_handler(sanitized_uri) do
-        @connection.start
+        connection.start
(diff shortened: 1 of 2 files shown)


bliof:
@michaelklishin Does this refactoring look ok?

michaelklishin:
@bliof I had basically the same idea just a couple of days ago. I will take a look later in the week but this looks promising, extensible, and very predictable for both existing and future users.

michaelklishin:
:+1:

parhs:
If using this with  threaded server like puma:
Thread pool is not fixed and pool can be trimmed wouldn't this end up having open channels that cannot be closed.
Has anyone tested it?

michaelklishin:
If a thread in the pool is reclaimed one way or another, all objects instantiated in that thread should be garbage collected eventually. This does not close channels on the RabbitMQ end, which can in theory produce a channel leak.

Use [per-vhost and per-user limits](https://rabbitmq.com/production-checklist.html#limits) to protect your clusters against such leaks. Monitoring should be able to answer your question
but it entirely depends on what thread pool implementation is used, and what kind of guarantees and controls it provides.

PR #200, opened 2016-03-03
Use stdlib JSON instead of multi_json gem

seamusabshere:


michaelklishin:
Why is this an improvement?

seamusabshere:
http://www.mikeperham.com/2016/02/09/kill-your-dependencies/

> So many gems declare a dependency on json, oj, multi_json, or yajl-ruby. There’s so many ossified layers of cruft around JSON processing that only one course of action makes sense: remove it all. JSON has been in the stdlib since 1.9

michaelklishin:
I'm sorry but one blog post is not a significant enough argument for me. It's not unreasonable to use YAJL over the json gem. `multi_json` is also a really unobtrusive dependency.

michaelklishin:
> A Rubygem should never use anything but Net::HTTP internally!

Oh yeah, the hard and fast rules from opinionated Ruby developers who know better. thanks, I stop reading right there and close this PR. Thank you for your time.

seamusabshere:
@michaelklishin 

i agree that Mike's post is very strongly (over-?)stated

but i would apply occam's razor here - why include a dependency if it's in the stdlib?

michaelklishin:
Because json gem is sufficient for many but not all.

Hutch has only 4 runtime dependencies: Bunny/March Hare, a Ruby HTTP API client, MultiJSON, ActiveSupport. It's not 39, and while the latter two could probably be eliminated with some effort and sacrifices, I find them to be fairly unobtrusive. To my knowledge the number of dependencies Hutch has has never been an issue for its users.

PR #201, opened 2016-03-18
Addresses some code duplication and tightens up tests in Hutch::CLI

rthbound:
- Consolidates some code in the CLI around aborting when
    required files don't exist
  - Tests the messaging when CLI encounters SystemExit


michaelklishin:
Thank you!

PR #202, opened 2016-03-20
Addressing complexity around CLI's option parsing

rthbound:
Refactors the CLI's options parsing to a CLIConfig module to reduce the CLI class's complexity.

### DIFF
--- a/lib/hutch/cli.rb
+++ b/lib/hutch/cli.rb
@@ -3,11 +3,12 @@ require 'optparse'
 require 'hutch/version'
 require 'hutch/logging'
 require 'hutch/exceptions'
-require 'hutch/config'
+require 'hutch/cli_config'
 
 module Hutch
   class CLI
     include Logging
+    include CLIConfig
 
     # Run a Hutch worker with the command line interface.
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you for your time. I'm afraid I don't see this as a simplification: it replaces a single block of code where all steps are highly related with a composed method. That sounds good in theory but now CLI option definition is harder to locate, there are 2 CLI modules, and a few methods not related to CLI argument parsing end up in the new module where I don't think they belong.

rthbound:
@michaelklishin: No worries, I enjoyed playing the CodeClimate game (2.9 -> 3.5). :wink:

I agree with your judgement. Thanks for taking the time to review.

rthbound:
@michaelklishin I meant to ask: should the methods in Hutch::CLI (besides `#run`) have private visibility? I see where `Hutch::CLI#parse_options` is used in tests, but I don't expect any of the other methods should be part of the public API.

michaelklishin:
Some methods in `Hutch::Client` can be made private (or protected, which I tend to use because you never know why someone might need to override your stuff).

PR #203, opened 2016-03-21
Reduce complexity of Hutch::Broker

rthbound:
- Adds an `#unbind_redundant_bindings` method to break up
    the complex `#bind_queue` method.
  - Adds a `#log_publication` method to break up the complex
    `#publish` method.

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -151,20 +151,24 @@ module Hutch
       results
     end
 
+    # Find the existing bindings, and unbind any redundant bindings
+    def unbind_redundant_bindings(queue, routing_keys)
+      return unless http_api_use_enabled?
+
+      bindings.each do |dest, keys|
+        next unless dest == queue.name
+        keys.reject { |key| routing_keys.include?(key) }.each do |key|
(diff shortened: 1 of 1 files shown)


lib/hutch/broker.rb
```diff
@@ -151,20 +151,24 @@ def bindings
       results
     end
 
+    # Find the existing bindings, and unbind any redundant bindings
+    def unbind_redundant_bindings(queue, routing_keys)
+      return unless http_api_use_enabled?
+
+      bindings.each do |dest, keys|
```
rthbound: Previously we iterated over bindings twice: once to filter, and a second time to do our work. Granted, the second time would likely iterate over a subset of bindings. Better still to iterate once if possible, skipping those we would've filtered and doing the necessary work otherwise.

rthbound:
@michaelklishin how would you feel about carving a `Publisher` class out of `Broker`?

I don't see any other ways to reduce its overall complexity.

michaelklishin:
@rthbound I feel positively about splitting `Hutch::Broker` but we need to keep things backwards compatible one way or another.

PR #204, opened 2016-03-21
Carves out a new Hutch::Publisher from Hutch::Broker

rthbound:
- Brings in `Hutch::Publisher` to do the publishing work for `Hutch::Broker`.
- Refactors methods specific to publishing to the new `Hutch::Publisher` class.

I believe these changes will be backwards compatible.

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -1,7 +1,7 @@
 require 'carrot-top'
-require 'securerandom'
 require 'hutch/logging'
 require 'hutch/exceptions'
+require 'hutch/publisher'
 
 module Hutch
   class Broker
@@ -51,6 +51,7 @@ module Hutch
       open_connection!
       open_channel!
       declare_exchange!
+      declare_publisher!
     end
 
     def open_connection
@@ -97,6 +98,10 @@ module Hutch
 
     def declare_exchange!(*args)
       @exchange = declare_exchange(*args)
+    end
+
+    def declare_publisher!
+      @publisher = Hutch::Publisher.new(connection, channel, exchange, @config)
     end
 
     # Set up the connection to the RabbitMQ management API. Unfortunately, this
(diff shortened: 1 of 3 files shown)


lib/hutch/broker.rb
```diff
@@ -215,41 +216,8 @@ def nack(delivery_tag)
...
-
-      response = exchange.publish(payload, {persistent: true}.
-        merge(properties).
-        merge(global_properties).
-        merge(non_overridable_properties))
-
-      channel.wait_for_confirms if @config[:force_publisher_confirms]
-      response
+    def publish(*args)
+      publisher.new(connection, channel, exchange, @config).publish(*args)
```
michaelklishin: Do we really want to instantiate a new publisher for every message published? There is no reason to do so since the connection, channel, exchange, and configuration are very much static. Connection recovery will take care of reviving the connection and channel after connection failures.

rthbound: Fair point. Refactored to avoid instantiating publisher for every message.

michaelklishin: Hm, I believe `publisher.new(…)` is still used in the `Hutch::Broker#publish` method?

rthbound: It should only be used on the first publication.

`publisher.publish(*args)` calls the `Broker#publisher` method which just returns `@publisher` if it's set. Otherwise `Broker#publisher` sets and returns `@publisher`.

michaelklishin: If we have `@publisher` initialised and its state is very much static, why do we need to do that? Why not instantiate it once (as we do) and use it?

rthbound: Well, you'll want to at least wait to initialise the publisher until connection, channel, and exchange are all set.

My thoughts were toward initialising the publisher object right before publishing, using memoization to ensure it's only done once. It sounds like you'd prefer to avoid memoization, and to initialise the publisher immediately after setting those properties.  I'll make those changes now (and override my memoization changes)

rthbound: @michaelklishin by calling `#declare_publisher!` just after `#declare_exchange!`, we get the behavior your asking for.

However, there are tests for publishing without a connection that break with this change. These tests work by calling `#publish` without having called `#connect`. Without calling `#connect`, `@publisher` remains nil, and so we get `undefined method #publish for nil:NilClass` rather than a Hutch::PublishError.

To repair this test, I think we'd need to modify the test so that #connect gets called but obtains an invalid connection (rather than no connection) **-or-** we'd need to ensure we wrap `Broker#publish` with `#with_connection_error_handler` and be sure to raise the appropriate error, rather than letting `NoMethodError` be raised.  IMO the memoization is much cleaner, but I'll do whatever you suggest.  What do you suggest?

rthbound: @michaelklishin it is easy enough to repair the broken tests with:
```ruby
before { broker.set_up_amqp_connection; broker.disconnect }
```

Changes coming momentarily.

michaelklishin:
Thank you!

PR #205, opened 2016-03-21
use true null logger for specs

kwstannard:
This will truly silence all logging from Hutch::Logging.logger


michaelklishin:
:+1:

PR #206, opened 2016-03-21
allow additional worker setup

kwstannard:
### Why
I needed to create a per-queue dead letter handling system and, judging by some of the old issues, Hutch would prefer to stay out of that business. So I needed a way to hook into Hutch during worker setup to do the dead letter stuff outside of Hutch.

### What
- Added an array of objects that will be called at the start of `worker#run` to Hutch::Config
- Refactored out the queue setup stuff to clear up the worker. it is now the first item of setup_procs.

### Usage
```
>> Hutch::Config.setup_procs << Proc.new { puts "HELLO WORLD" }

>> Hutch::Config.setup_procs << other_object_that_responds_to_call
```

### Other stuff
This isn't quite in a great state. The MessageHandler class is basically a dump of functionality where I realized I should probably open a PR and see if this had any legs before I put too much time into the refactor. If this basic idea is good I would be happy to polish this up more.

I also have another branch that builds on the worker cleanup in this and gets worker down to 51 fully tested lines. [worker.rb](https://github.com/CasperSleep/hutch/blob/worker_refactor/lib/hutch/worker.rb) [worker_spec.rb](https://github.com/CasperSleep/hutch/blob/worker_refactor/spec/hutch/worker_spec.rb)

### DIFF
--- a/lib/hutch/cli.rb
+++ b/lib/hutch/cli.rb
@@ -85,7 +85,7 @@ module Hutch
     # gracefully (with a SIGQUIT, SIGTERM or SIGINT).
     def start_work_loop
       Hutch.connect
-      @worker = Hutch::Worker.new(Hutch.broker, Hutch.consumers)
+      @worker = Hutch::Worker.new(Hutch.broker, Hutch::Config[:setup_procs])
       @worker.run
       :success
     rescue ConnectionError, AuthenticationError, WorkerSetupError => ex
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -1,4 +1,5 @@
 require 'hutch/error_handlers/logger'
+require 'hutch/setup/queues'
 require 'erb'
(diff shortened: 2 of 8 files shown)


michaelklishin:
I think this has two unrelated changes. Running setup steps is separate from message [pre-]processing.

And if we are going to have message pre-processors we might as well make it possible to register your own (also as callables?) Please put that in another PR.

kwstannard:
@michaelklishin I am not sure what you are asking. Do you want me to move all the functionality back into the worker? What is message pre-processing?

michaelklishin:
What `MessageHandler` does in your this PR is message pre-processing, before it is passed on to the actual consumer.

kwstannard:
Pre-processing was a dependency of queue setup in the original worker code. All I did was cut paste code from worker into Setup::Queue and MessageHandler and add the new config var.

PR #207, opened 2016-03-23
allow for functions to run before main loop

kwstannard:
Basically this pull request but without any refactoring: https://github.com/gocardless/hutch/pull/206

### DIFF
--- a/lib/hutch/worker.rb
+++ b/lib/hutch/worker.rb
@@ -10,9 +10,10 @@ module Hutch
 
     SHUTDOWN_SIGNALS = %w(QUIT TERM INT)
 
-    def initialize(broker, consumers)
+    def initialize(broker, consumers, setup_procs)
       @broker        = broker
       self.consumers = consumers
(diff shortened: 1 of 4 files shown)


lib/hutch/worker.rb
```diff
@@ -20,6 +21,7 @@ def initialize(broker, consumers)
     # never returns.
     def run
       setup_queues
+      setup_procs.each(&:call)
```
michaelklishin: Why does the worker invoke those callables? Is it to make sure the queues are set up first?

kwstannard: Because the worker sets up the queues, I figured that the other setup stuff should go here as well.

kwstannard:
@michaelklishin if I were to move the queue setup back into its own class for the setup loop, would you like me to move the setup loop out into the CLI?

michaelklishin:
@kwstannard I don't think we need to move queue setup to a separate class.

PR #210, opened 2016-04-17
Silence Bunny Logger in tests

kwstannard:
Set client_logger in broker_spec to log to /dev/null.

I also got rid of deep_copy. The config should just be the default,
which means a call to Hutch::Logger.initialize will suffice.

Finally, I made the expected error in broker_spec more specific.

### DIFF
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -121,11 +121,5 @@ module Hutch
         get(attr)
       end
     end
-
-    private
-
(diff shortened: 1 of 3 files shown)


kwstannard:
There is a failure on JRuby. Looking into it

kwstannard:
Handling the march hare error the same way as the bunny error will technically break the api, so I will just revert the change to the error expectation.

PR #212, opened 2016-04-23
Validate partial doubles

kwstannard:
RSpec allows you to validate that mocked methods exist on classes, so I thought I would turn it on and fix what popped up.

### DIFF
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -1,4 +1,6 @@
 require 'hutch/error_handlers/logger'
+require 'hutch/tracers'
+require 'hutch/serializers/json'
 require 'erb'
 require 'logger'
 
@@ -39,7 +41,7 @@ module Hutch
 
     def self.initialize(params = {})
       @config = default_config
-      @config.merge(env_based_config).merge(params)
+      @config.merge!(env_based_config).merge!(params)
       define_methods
       @config
     end
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -129,10 +129,6 @@ module Hutch
       ALL_KEYS.select { |attr| ENV.key?(key_for(attr)) }
     end
 
-    def self.reset!
-      @config = initialize
-    end
-
     def self.get(attr)
       check_attr(attr)
       user_config[attr]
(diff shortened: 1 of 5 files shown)


michaelklishin:
Can we please merge the RSpec and error handler changes in two separate PRs? Thank you.

kwstannard:
okay, I will probably do that sometime this weekend

michaelklishin:
@kwstannard thank you.

michaelklishin:
Now that #220 is submitted, can we please restore error handlers here and focus on test doubles?

kwstannard:
Sorry it took so long, the ENV PR really changed a lot of things around.

michaelklishin:
I have a test failure with this PR:

```
Failures:

  1) Hutch::ErrorHandlers::Airbrake#handle logs the error to Airbrake
     Failure/Error: expect(::Airbrake).to receive(:notify_or_ignore).with(ex, message)
       Airbrake does not implement: notify_or_ignore
     # ./spec/hutch/error_handlers/airbrake_spec.rb:30:in `block (3 levels) in <top (required)>'
```

lib/hutch/config.rb
```diff
@@ -189,14 +190,15 @@ def self.convert_value(attr, value)
...
+          get(key)
+        end
 
-      if method =~ /=$/
-        set(attr, args.first)
-      else
-        get(attr)
+        define_singleton_method("#{key}=") do |val|
+          set(key, val)
+        end
```
olleolleolle: This is a _great_ part of this change.

With this, we can wrap the built-in `define_singleton_method(key)` into a YARD-documentable meta-programming method – which creates these `attr_accessor`-style methods using the above code.

[The YARD author uses `@macro` to great effect, in this example](https://groups.google.com/forum/#!topic/yardoc/HIUoeATOjWA).

That allows us to simplify the YARD "settings" extension to only cover collecting the `string_setting` and friends.

Question: Does this version of `.define_methods` want to allow the `params` given to `.initialize` to influence which get/set convenience methods are available?

kwstannard: Yes. Anything passed in params will be defined as a method.

michaelklishin:
Thank you!

kwstannard:
you're welcome

PR #216, opened 2016-04-27
WIP: YARD settings, docs

olleolleolle:
This PR introduces config for [Yardoc](http://yardoc.org/), a system to render API docs.

  - ignore rendered output `doc/` folder (we can publish it just after we render it)
  - decide to use Markdown to style comments
  - avoid a yardoc warning by expanding something that looked like an internal link to it

**Update**: Since Open Source is amazing, the YARD output is already automatically rendered by http://www.rubydoc.info/gems/hutch - so we can merely author prettier docs for it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ Gemfile.lock
 .ruby-gemset
 .idea
 coverage
+.yardoc
+doc
--- a/.yardopts
+++ b/.yardopts
@@ -1,0 +1,2 @@
+-m markdown
+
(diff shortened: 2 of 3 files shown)


michaelklishin:
Should we add YARD as a development dependency, too?

olleolleolle:
(Haha, I just moved to do that, and then this was merged.)

```ruby
gem.add_development_dependency 'yard', '~> 0.8'
```

olleolleolle:
If we're clever, we can perhaps define a Yard macro to render those config options for us, from code docs.

PR #219, opened 2016-04-29
Docs: Hutch::Config

olleolleolle:
In order to help users configure Hutch, try to tell the configuration story using YARD docs. This PR now produces docs.

screenshot 2016-04-29 06 55 13

Using a YARD processor, the names, types and default values can be presented in the documentation:

screenshot 2016-04-30 15 14 21

(Using the power of a CSS class, we could style that to be legible.)

### DIFF
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -15,29 +15,29 @@ module Hutch
   #   HUTCH_PUBLISHER_CONFIRMS=false hutch
   module Config
     require 'yaml'
-    STRING_KEYS = Set.new
-    NUMBER_KEYS = Set.new
-    BOOL_KEYS = Set.new
+    @string_keys = Set.new
+    @number_keys = Set.new
+    @boolean_keys = Set.new
     @settings_defaults = {}
 
     # Define a String user setting
     # @!visibility private
     def self.string_setting(name, default_value)
-      STRING_KEYS << name
+      @string_keys << name
       @settings_defaults[name] = default_value
     end
 
     # Define a Number user setting
     # @!visibility private
     def self.number_setting(name, default_value)
-      NUMBER_KEYS << name
+      @number_keys << name
       @settings_defaults[name] = default_value
     end
 
     # Define a Boolean user setting
     # @!visibility private
     def self.boolean_setting(name, default_value)
-      BOOL_KEYS << name
+      @boolean_keys << name
       @settings_defaults[name] = default_value
     end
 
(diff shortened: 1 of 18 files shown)


lib/hutch/config.rb
```diff
@@ -5,42 +5,75 @@
...
-                   daemonise
-                   publisher_confirms
-                   force_publisher_confirms
-                   enable_http_api_use
-                   consumer_pool_abort_on_exception).freeze
-
-    ALL_KEYS = (BOOL_KEYS + NUMBER_KEYS + STRING_KEYS).map(&:to_sym).freeze
+    STRING_KEYS = Set.new
+    NUMBER_KEYS = Set.new
+    BOOL_KEYS = Set.new
```
olleolleolle: TODO: Make these not be defined as constants.

lib/hutch/config.rb
```diff
@@ -5,42 +5,75 @@
...
+
+    boolean_setting :mq_tls, false
+    boolean_setting :mq_verify_peer, true
+    boolean_setting :mq_api_ssl, false
+    boolean_setting :autoload_rails, true
+    boolean_setting :daemonise, false
+    boolean_setting :publisher_confirms, false
+    boolean_setting :force_publisher_confirms, false
+    boolean_setting :enable_http_api_use, true
+    boolean_setting :consumer_pool_abort_on_exception, false
```
olleolleolle: Perhaps these DSL-like methods could help set up the default_settings in some way.

lib/yard-settings/handler.rb
```diff
@@ -0,0 +1,48 @@
...
+  handles method_call :string_setting
+  handles method_call :number_setting
+  handles method_call :boolean_setting
+
+  namespace_only
+
+  def process
+    name = statement.parameters.first.jump(:tstring_content, :ident).source
+    object = YARD::CodeObjects::MethodObject.new(namespace, name)
+    register(object)
```
olleolleolle: This is the "render each of the settings as a method on the Module" bit. Perhaps this is over-reaching, as the non-ENV-settable settings (which are also available as method calls via method_missing) will not be represented here.

Example of what the current output is:

screenshot 2016-04-30 15 18 43

templates/default/layout/html/settings.erb
```diff
@@ -0,0 +1,17 @@
...
+    <tr>
+    <th><tt><%= task.full_path %></tt></th>
+    <% if task.full_path == 'default' %>
+    <th><%= task.tag(:task_deps).text %></th>
+    <% else %>
+    <td><%= task.docstring %></td>
+    <% end %>
+    </tr>
+  <% end %>
+</table>
```
olleolleolle: This file should go.

michaelklishin:
This is awesome, thank you @olleolleolle! I'd be quite happy to merge the same kind of PR for Bunny ;)

olleolleolle:
@michaelklishin Thanks for merging - I believe there are a few clean-ups that can be achieved. I'll be PRing on that account soon.

lib/yard-settings/handler.rb
olleolleolle: @michaelklishin This is the place where I'd like your input on also creating "`read_timeout=`" as well (perhaps with the "Found using `method_missing`" addition, too?

PR #220, opened 2016-04-30
ErrorHandler dependencies.

kwstannard:
Paging error handler authors: @michaelklishin @natesalisbury @danielfarrell @karls

We last left off with this: https://github.com/gocardless/hutch/pull/212

The work I have done so far splitting out gems is not super important right now as there should probably be some sort of discussion.

- Versioning: What is the best way to do this? Something like rspec-rails where it is somewhat tied to the parent gem version? Completely independent version?
- Should hutch itself go in a subfolder?
- Is this too much? Should we just delete the error handlers and have end users implement their own? Should we just put them in an examples folder that is not actually in the gem?


michaelklishin:
I don't think we should delete them. Having a separate gem is OK. It can be versioned independently from Hutch IMO.

danielfarrell:
Agreed, separate gems is probably the way to go. We can split them into
their own repos as well if you want them out of the hutch repo.

kwstannard:
How about some pros-cons
Separate:
+ less complicated. Can just use boilerplate for CI.
+ easier to drop support for unused gem.
Single:
+ harder to ignore when a vendor specific gem is out of date. 

If we aren't having any sort of version parity then I would lean towards separate repos.

michaelklishin:
@kwstannard I'm not against using separate repos eventually. Just prefer doing breaking changes like this one at a time.

If I can get a blessing from @gocardless engineers, this is good to go with me.

kwstannard:
Okay, we probably want to just close this PR if you decide on seperate repos.

PR #229, opened 2016-05-30
Message preprocessor and worker refactoring

kwstannard:
Thought I would finish this refactoring up because it is an interesting problem space and I liked the direction it was taking. You mentioned having a changeable preprocessor before, so I added in a config var that you can set to change the pre-processing.

The main thrust of refactoring is that the worker class is one of the central parts of Hutch, and as such it was made as simple as possible. Loop logic, queue setup logic, and message pre-processing logic was moved out and tested.

One concern is that some public functions in worker were moved to other classes and some of those were further made private. My thought was that they looked like they were only public for testing purposes, but you would know better than me. If this is merged you might want to make the next release a major version change just in case.

### DIFF
--- a/lib/hutch/main_loop.rb
+++ b/lib/hutch/main_loop.rb
@@ -1,0 +1,41 @@
+require 'hutch/logging'
+
+module Hutch
+  class MainLoop
+    include Logging
+
+    SHUTDOWN_SIGNALS = %w(QUIT TERM INT)
+
+    def self.loop_until_signaled
+      new.loop_until_signaled
+    end
+
+    def loop_until_signaled
+      self.sig_read, self.sig_write = IO.pipe
+
+      register_signal_handlers
+      wait_for_signal
+
+      sig = sig_read.gets.strip.downcase
+      logger.info "caught sig#{sig}, stopping hutch..."
+    end
+
+    private
+
+    attr_accessor :sig_read, :sig_write
+
+    def wait_for_signal
+      IO.select([sig_read])
+    end
+
+    def register_signal_handlers
+      SHUTDOWN_SIGNALS.each do |sig|
+        # This needs to be reentrant, so we queue up signals to be handled
+        # in the run loop, rather than acting on signals here
+        trap(sig) do
+          sig_write.puts(sig)
+        end
+      end
+    end
+  end
+end
--- a/lib/hutch.rb
+++ b/lib/hutch.rb
@@ -12,6 +12,7 @@ require 'hutch/version'
 require 'hutch/error_handlers'
 require 'hutch/exceptions'
 require 'hutch/tracers'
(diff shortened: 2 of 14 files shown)


lib/hutch/main_loop.rb
```diff
@@ -0,0 +1,41 @@
+require 'hutch/logging'
+
+module Hutch
+  class MainLoop
```
kwstannard: Most of MainLoop is inspired by this PR on sidekiq: https://github.com/mperham/sidekiq/pull/761

Not sure if loop is the correct name for this anymore, but I am struggling to think of a better one.

kwstannard:
Looks like Travis is having trouble with the SIGQUIT signal test. `bundle exec rake` passes for me locally, so maybe this is a difference in linux vs osx?

kwstannard:
It appears that some changes happened to Hutch::Config since last time I worked on this and client_logger was deleted.

michaelklishin:
`MessagePreprocessor` does more than "preprocessing". I am not convinced we need a separate class just to declare queues. Sorry but the value of mock-only tests in this case is not obvious to me either.

I see no reason to merge this PR.

kwstannard:
I take it since this is still open, that you are open to something happening.

Fair point on not being convinced, not everyone believes in SOLID and I shouldn't expect that. I went on this path because I couldn't figure out what was happening in the worker class without pulling it apart. There was too much happening in there. Once I pulled it apart I figured I might as well clean it up a bit, update the loop logic, and open a pull request.

As for mock only tests I am assuming you are talking about the new worker spec. I felt it was necessary to spec the order of operations. Without that test there was nothing in worker spec since everything else that was originally there was moved to the message pre-processing spec or the queue spec. I can see a case for an integration test but there weren't any integration tests in the first place, so I did not make the test suite worse in this PR and in fact added non-mock tests for crucial components like the loop sequence which was completely untested before.

Can you just tell me what lines you think are pre-processing? Is it the Message class itself?

michaelklishin:
This is open because I wanted to wait to see if others feel otherwise and because the signal handling parts are probably worth extracting.

I had something closer to middleware in mind for "pre-processing." If it's just doing the same thing, I'd just leave it in `Hutch::Worker`.

kwstannard:
Okay. Pre-processing as a middleware insert point makes sense to me. What would you think about a rack style chain with each middleware calling the next until it hits the consumer.process method? It might even be compatible with certain pre-existing rack middleware if it uses the same api. Done right, it would replace the error handlers and tracers as well.

I will definitely remove the config addition, and then we can wait for some other opinions on the rest.

michaelklishin:
@kwstannard let's not replace error handlers and tracers yet. Tracers should be trivial to replace, error handlers can be much trickier.

Please submit the signal handling loop in a separate PR.

PR #230, opened 2016-06-03
extract process holding logic

kwstannard:
Pulled loop logic out, updated it to use IO.select, and removed cruft
methods.


kwstannard:
I am working with travis support to resolve the failures.

kwstannard:
Okay figured it out.

michaelklishin:
@kwstannard thank you!

PR #231, opened 2016-06-06
Send payload in Sentry error handler

matt-thomson:
When debugging errors in Hutch consumers, it's useful to be able to see the problematic message payload.  The error handlers for Airbrake and Honeybadger both send the payload as part of the notification, but the Sentry error handler does not.

This PR adds the payload in the `extra` information for Sentry, meaning that it'll show up alongside the error.


michaelklishin:
We already do this for Airbrake and Honey Badger, so this makes perfect sense. Thank you.

hmarr:
👍 from me!

matt-thomson:
Thanks @michaelklishin - please could you cut a release with this fix in place?

michaelklishin:
@matt-thomson will do in a bit.

PR #236, opened 2016-06-23
Don't try to register unsupported Signals

tobmatth:
Signal QUIT is not support under Windows


michaelklishin:
Thank you!

tobmatth:
Wow, that was fast :) Thanks!

PR #237, opened 2016-06-29
Worker: Remove unused constant

olleolleolle:
In #236 it became clear to me that this other same-name constant `SHUTDOWN_SIGNALS` was not seeing any use in Worker. (Waiter has the right one.)

This PR removes it.


michaelklishin:
Thank you!

PR #238, opened 2016-07-05
Allows to pass all message properties to error handlers.

gottfrois:
Message properties can include valuable message informations such as
correlation id, message type, app specific headers and many more.
When handling consumer process errors, we should be able to make decisions
based on any message properties.

Note: This is a breaking change in case people created their own error handlers.

### DIFF
--- a/lib/hutch/error_handlers/airbrake.rb
+++ b/lib/hutch/error_handlers/airbrake.rb
@@ -6,7 +6,8 @@ module Hutch
     class Airbrake
       include Logging
 
-      def handle(message_id, payload, consumer, ex)
+      def handle(properties, payload, consumer, ex)
+        message_id = properties.message_id
         prefix = "message(#{message_id || '-'}): "
         logger.error prefix + "Logging event to Airbrake"
         logger.error prefix + "#{ex.class} - #{ex.message}"
--- a/hutch.gemspec
+++ b/hutch.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |gem|
   end
   gem.add_runtime_dependency 'carrot-top', '~> 0.0.7'
   gem.add_runtime_dependency 'multi_json', '~> 1.11.2'
-  gem.add_runtime_dependency 'activesupport', '>= 3.0'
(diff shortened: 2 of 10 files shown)


lib/hutch/error_handlers/airbrake.rb
```diff
@@ -6,7 +6,8 @@ module ErrorHandlers
     class Airbrake
       include Logging
 
-      def handle(message_id, payload, consumer, ex)
+      def handle(properties:, payload:, consumer:, ex:)
```
michaelklishin: I'm not familiar with this syntax.

michaelklishin: OK, so these are required keyword arguments that are available in Ruby 2.0.

gottfrois: correct, I looked at your travis definition and seems we only care about ruby >= 2.0

```
  - "2.3.0"
  - "2.2"
  - "2.1"
  - "2.0"
  - "jruby-9.0.0.0"
```

gottfrois:
The spec fails because of latest activesupport release:

```
Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
```

A choice needs to be made here, whether we want to support older ruby versions or use latest activesupport.

lib/hutch/error_handlers/airbrake.rb
michaelklishin: Bunny only supports 2.0+ in recent versions, so that's fine.

michaelklishin:
@gottfrois let's pin to a `4.x` version for now.

michaelklishin:
@gottfrois perhaps we want to depend on `~ 4.2` does that version also not support Ruby 2.0?

gottfrois:
> Ruby 2.0 doesn’t have built-in support for required keyword arguments. 

My bad, ruby 2.0 don't support `(foo:, bar:)`. Will push a fix.

michaelklishin:
OK, even though this is a breaking change in the error handler API, it makes sense. I'll update change log to highlight this.

PR #239, opened 2016-07-07
Draw consumers info table

andrey-skat:
Add cli parameter to show consumers and they routes and queues like rails routes.

`$ hutch --consumers`
```
Consumer                     | Queue name           | Route
--------------------------------------------------------------------------
Admin::AdminApprovedConsumer | admin:admin_approved | admin.approved
UserCreatedConsumer          | user_created         | user.created
UserDeletedConsumer          | user_deleted         | user.deleted
UserModifiedConsumer         | user_modified        | user.updated, user.deleted
```


lib/hutch/cli.rb
```diff
@@ -189,6 +189,12 @@ def parse_options(args = ARGV)
           Hutch::Config.pidfile = pidfile
         end
 
+        opts.on('--consumers', 'Show consumers information') do
+          load_app
```
andrey-skat: Initially I want to make `Dir.glob` through require paths and don't load rails app. But it adds separate load consumers logic.

michaelklishin:
I personally think this should be a part of a separate tool. Not how useful this feature is either, no one has requested it to date.

michaelklishin:
"route" is not descriptive, it should be "routing keys" but even then it would be confusing because consumers do not have routing keys, bindings (and messages, of course) do.

lib/hutch/cli.rb
olleolleolle: Perhaps the description of this option could be "Print consumers information and quit"?

olleolleolle:
@andrey-skat I got carried away. https://gist.github.com/olleolleolle/e130dacfba3ff59b8d4a8e9b524b4c91 - does this do the same?

olleolleolle:
I guess this tool answers the question "Does Hutch find and load my Consumer?"

michaelklishin:
Then the flag should at least be `--list-consumers`? Or the answer is "improve consumer logging instead"?

andrey-skat:
> "route" is not descriptive, it should be "routing keys" but even then it would be confusing because consumers do not have routing keys, bindings (and messages, of course) do.

Yes consumers don't have routing keys, but they bind them to queue.

This table helps overview what queues (and routes) we consume.

lib/hutch/cli.rb
```diff
@@ -189,6 +189,12 @@ def parse_options(args = ARGV)
           Hutch::Config.pidfile = pidfile
         end
 
+        opts.on('--consumers', 'Print consumers information and exit') do
```
andrey-skat: Description updated

lib/hutch/cli/consumers_inspector.rb
```diff
@@ -0,0 +1,44 @@
...
+            route: consumer.routing_keys.to_a.join(', ')
+          }
+        end
+      end
+
+      def draw_table(consumers)
+        consumers_table = []
+        class_width, queue_width, route_width = widths(consumers)
+
+        consumers_table << "#{'Consumer'.ljust(class_width)} | #{'Queue name'.ljust(queue_width)} | Routing keys"
```
andrey-skat: "Route" changed to "Routing keys"

andrey-skat:
> @andrey-skat I got carried away. https://gist.github.com/olleolleolle/e130dacfba3ff59b8d4a8e9b524b4c91 - does this do the same?

I don't think that it should has so many instance variables and attr_readers.

I can extract some code from `draw_table` if it seems big. But I think is too early to overcomplicate this simple task.

michaelklishin:
I'll leave it to others to judge how useful this feature is. To me it makes more sense to improve logging instead.

michaelklishin:
No feedback in almost 1 full month. Should this be closed in favour of logging improvements?

PR #241, opened 2016-08-06
Initialize Hutch::Config right after loading

### LINKED ISSUE #240: Logger resets Hutch configuration
We are using Hutch within an Rails application and we have a problem with producers if we doesn't setup the logger in the initializer.
The following is not working:
`Hutch::Config.set(:mq_exchange, 'name')`

This works:
```
Hutch::Logging.logger = Rails.logger
Hutch::Config.set(:mq_exchange, 'name')
```

The problems seems to be [line 14 in logging.rb](https://github.com/gocardless/hutch/blob/v0.22.1/lib/hutch/logging.rb#L14). If you set the logger in the initializer the whole method `setup_logger` is not called.
I don't know if it's needed to load the default setup or if you could merge an existing configuration in `config.rb initialize`.

michaelklishin:
Fixes #240.

### DIFF
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -264,3 +264,5 @@ module Hutch
     end
   end
 end
+
+Hutch::Config.initialize


olleolleolle:
Line 270 in Config in master now does this, right?
https://github.com/gocardless/hutch/blob/master/lib/hutch/config.rb#L270

michaelklishin:
Yes, this is now obsolete thanks to the merged PR from @olleolleolle.

PR #242, opened 2016-08-15
Set required Ruby version

sgerrand:
💁  The dependencies of this gem require a modern version of Ruby (see the following stderr output).

Specifying the required Ruby version in the configuration should help consumers using deprecated Rubies by failing fast during `gem install`, rather than during `bundle install`.

```
Gem::InstallError: amq-protocol requires Ruby version >= 2.0.

Gem::InstallError: mime-types-data requires Ruby version >= 2.0.

Gem::InstallError: tins requires Ruby version >= 2.0.
```


michaelklishin:
Thank you very much!

PR #243, opened 2016-09-12
Depend on activesupport >= 4.0 on Ruby versions greater than 2.3

olleolleolle:
In order to use latest Rails, allow >= 4.0 of activesupport on Ruby versions larger than 2.3.


michaelklishin:
This means [dropping support for Ruby versions older than 2.3](https://travis-ci.org/gocardless/hutch/jobs/159314633) (by default, at least).

I cannot merge this without a blessing from @gocardless engineers.

olleolleolle:
@michaelklishin Thank you for the explicit support notice. The diligence is much appreciated.

olleolleolle:
@michaelklishin Now, all versions build green, with a conditional in the gemspec.

michaelklishin:
Thank you!

PR #247, opened 2016-09-21
Bump gem multi_json 1.12.1

olleolleolle:
This PR updates the multi_json gem in the gemspec.

- [CHANGELOG](https://github.com/intridea/multi_json/blob/master/CHANGELOG.md)


andrey-skat:
Why you need so strict dependency?

`'multi_json', '~> 1.12'` should not break anything and not block other gems to install higher versions.

olleolleolle:
@andrey-skat Good question! Please add a PR with your idea!

michaelklishin:
No objections to @andrey-skat's idea from me.

andrey-skat:
Added PR https://github.com/gocardless/hutch/pull/255

Issue #244, opened 2016-09-20
Dynamically picking version of ActiveSupport won't work as expected


jensnockert:
https://github.com/gocardless/hutch/commit/814c1e852c044e9d7278490115841f5870886efc sets the ActiveSupport version at Gem build time, so will pick dependency depending which RUBY_VERSION it was built on, not what RUBY_VERSION it is installed on. :/

I'm not sure what the best way to actually solve this issue is. I'll try to catch @olleolleolle and send in a proper.

jensnockert:
An extension like the following 'works', except if you use Bundler which won't pick up the dependency.

```ruby
require 'rubygems/dependency_installer.rb'

begin
  activesupport_version = (RUBY_VERSION >= '2.3' ? '>= 4.0' : '~> 4.0')

  installer = Gem::DependencyInstaller.new
  installer.install 'activesupport', activesupport_version
rescue
  exit(1)
end

f = File.open(File.join(File.dirname(__FILE__), "Rakefile"), "w")
f.write("task :default\n")
f.close
```

michaelklishin:
This Ruby version dependency stuff is depressing.

olleolleolle:
Let me express it with a `:-|` neutral emoji face. 

We're all in this together.

jensnockert:
https://github.com/bundler/bundler/issues/4960 is maybe relevant?

michaelklishin:
I hope this was addressed by a bunch of PRs in master that uses separate Gemfiles for CI and drops the version constraint from Active Support (which should be fine, this gem's needs are small).

jensnockert:
Yes.

PR #248, opened 2016-09-21
Gemspec: Move development deps to Gemfile

olleolleolle:
This PR moves the development dependencies out of the gemspec and into groups in the Gemfile.

It also moves the version-dependent activesupport dependency into separated Gemfiles, so that the Travis build matrix targets can choose which activesupport version is for them.

By trying to leverage Bundler's `platform` support, we can get away from weird hacks in the gemspec.

In the Travis build matrix each Ruby version gets its dependencies pointed out to it using the `gemfile:` travis.yml feature. A `gemfiles/` folder is supplied.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Gemfile.lock
 coverage
 .yardoc
 doc
+gemfiles/Gemfile.*.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,18 @@ before_install: gem i bundler
 matrix:
   include:
     - rvm: 2.3.0
+      gemfile: gemfiles/Gemfile.activesupport5
     - rvm: 2.2
+      gemfile: gemfiles/Gemfile.activesupport4
     - rvm: 2.1
+      gemfile: gemfiles/Gemfile.activesupport4
     - rvm: 2.0
+      gemfile: gemfiles/Gemfile.activesupport4
     - rvm: jruby-9.1.5.0
-      env: JRUBY_OPTS='--debug'
       jdk: oraclejdk8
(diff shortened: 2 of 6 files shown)


michaelklishin:
What's the benefit of this?

olleolleolle:
After this, we can make more Gemfiles, to be able to run them next to each other, in the test platform matrix in Travis.

"You get an old ActiveSupport, you get a newer one." 

All chosen via Gemfile versions, all appropriate to the versions we are testing.

olleolleolle:
In [travis.yml](https://github.com/gocardless/hutch/blob/master/.travis.yml#L10), we can include an env variable `BUNDLE_GEMFILE=gemfiles/Gemfile.activesupport-4` for the ones that need it.

And then, get rid of the weird `RUBY_VERSION > '2.3'` thing in the gemspec.

olleolleolle:
I could push on, in that direction, in this PR, to see if the changes bears out the promise.

Or, we could try taking it a step at a time.

michaelklishin:
Please take your idea as far as you feel necessary.

olleolleolle:
@michaelklishin Here's what the overview looks like when using the `gemfile` directive in the Travis build matrix https://travis-ci.org/gocardless/hutch/builds/161750953

olleolleolle:
@jensnockert Does this work around #244?

olleolleolle:
(Squashed to 1 commit: most learning Travis is best kept out of the long-term commit log.)

michaelklishin:
This no longer merges cleanly.

michaelklishin:
@olleolleolle thank you, I like this solution.

olleolleolle:
@michaelklishin Rebased. Thanks for the encouragement.

PR #252, opened 2016-09-21
Gemfile: guard and guard-rspec work only on mri_23

olleolleolle:
This PR wants to fix master so that [guard](https://github.com/guard/guard) and guard-rspec do not disturb other Travis build targets


olleolleolle:
@michaelklishin There, the different mri's have now built.

PR #260, opened 2016-10-19
Logging: remove redundant Hutch::Config.initialize

olleolleolle:
The last statement of the Config class is to run initialize. We don't need to do it twice.

### DIFF
--- a/lib/hutch/logging.rb
+++ b/lib/hutch/logging.rb
@@ -9,12 +9,12 @@ module Hutch
       end
     end
 
-    def self.setup_logger(target = $stdout)
+    def self.setup_logger
       require 'hutch/config'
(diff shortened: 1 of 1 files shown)


michaelklishin:
I'd like to do some more testing with various Hutch CLI scenarios before we can be confident that it's no longer needed.

michaelklishin:
Still needs some testing on my end. I likely won't have the time to do it next week, sorry.

olleolleolle:
@michaelklishin I'm not in a rush to have this land.

I logged #261 as a follow-up issue for this when I found it. It's a hen-and-egg of Config defaults and settings. (Colloquially described: The logger wants a level before a level has been resolved to a real setting value.)

PR #262, opened 2016-10-27
Add an Opbeat Tracer

olleolleolle:
This PR adds a new autoloaded Tracer implementation for Opbeat.

- [x] load it only if Opbeat is defined (this is taken care of just like newrelic_rpm is required in the file that is autoloaded)
- [x] ask @mikker if this is the way to use Opbeat.trace

### DIFF
--- a/lib/hutch/tracers/opbeat.rb
+++ b/lib/hutch/tracers/opbeat.rb
@@ -1,0 +1,37 @@
+require 'opbeat'
+
+module Hutch
+  module Tracers
+    # Tracer for Opbeat, which traces each message processed.
+    class Opbeat
+      KIND = 'messaging.hutch'.freeze
+
+      # @param klass [Consumer] Consumer instance (!)
+      def initialize(klass)
+        @klass = klass
+      end
+
+      # @param message [Message]
+      def handle(message)
+        ::Opbeat.transaction(sig, KIND, extra: extra_from(message)) do
+          @klass.process(message)
+        end.done(true)
+      end
+
+      private
+
+      def sig
+        @klass.class.name
+      end
+
+      def extra_from(message)
+        {
+          body: message.body.to_s,
+          message_id: message.message_id,
+          timestamp: message.timestamp,
+          routing_key: message.routing_key
+        }
+      end
+    end
+  end
+end
--- a/lib/hutch/tracers.rb
+++ b/lib/hutch/tracers.rb
@@ -2,5 +2,6 @@ module Hutch
   module Tracers
     autoload :NullTracer, 'hutch/tracers/null_tracer'
     autoload :NewRelic,   'hutch/tracers/newrelic'
+    autoload :Opbeat,     'hutch/tracers/opbeat'
   end
 end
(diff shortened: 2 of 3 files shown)


mikker:
Nice! I don't know Hutch – but I noticed one thing. You need to wrap the whole thing in a transaction. Translations have traces. In Rails for example, every action is wrapped in a Transaction, and the DB calls are traces.

```ruby
Opbeat.transaction(sig, KIND, extra: extra_from(message)) do
  @klass.process(message)
end.done(some_kind_of_result_eg_true)
```

Be aware that the Opbeat client might need to be started manually if running in another process than Rails (where it's started doing boot).

olleolleolle:
@mikker Is _this_ how to use trace/transaction?

lib/hutch/tracers/opbeat.rb
```diff
@@ -12,9 +12,11 @@ def initialize(klass)
 
       # @param message [Message]
       def handle(message)
-        ::Opbeat.trace(
-          sig, KIND, extra: extra_from(message)
-        ) { @klass.process(message) }
+        ::Opbeat.trace(sig, KIND, extra: extra_from(message)) {
+          ::Opbeat.transaction(sig, KIND, extra: extra_from(message)) do
```
mikker: Other way around 😄  `Transaction(Trace())`

olleolleolle:
@mikker Thanks for checking my work. Re-check?

lib/hutch/tracers/opbeat.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      private
+
+      def sig
+        @klass.class.name
+      end
+
+      def extra_from(message)
+        {
+          body: message.body.to_s,
```
olleolleolle: Consider: Make this _scrubbable_, somehow?

mikker: Scrubbable how? You don't need the trace, actually. Transaction automatically have one root trace, so adding another one is probably redundant. I can see how I could've let you to add one. Sorry.

olleolleolle: Thanks for the note on the trace, I'll remove that inner bit.

olleolleolle: The scrubbing feature is way outside the scope of this PR. (Use-cases include: Anonymize hospital person data, for instance. `Name=Jane Doe`, that kind of thing.)

olleolleolle:
@michaelklishin What's your preferred way to "only ever load this `if defined?(Opbeat)`"?

olleolleolle:
The JRuby seems to have had a hiccup on an unrelated test.

michaelklishin:
@olleolleolle `Kernel#autoload`? just a condition with a require? I don't have a strong opinion.

michaelklishin:
Thank you!

mikker:
Thank you for this! 👍  (cc @roncohen)

olleolleolle:
Go team! So many incremental baby-steps, so many helpful nudges and high-fives.

PR #263, opened 2016-11-18
Perform a VM-wide thread stack trace dump on USR2

olleolleolle:
This is a PR inspired by another project which had a neat section of signals to send to the worker, to have to do things, like "shed current work item, never to requeue it" or "print a debug backtrace".

### Inspiration

- [Kuryuk - Signals chapter of docs](http://kuyruk.readthedocs.io/en/latest/signals.html)
- [Mike Perham's Sidekiq's wiki on Signals](https://github.com/mperham/sidekiq/wiki/Signals)

### Questions

- What signals should the Worker process handle?
- Can you golf the control structure exception away? (See - this is me feeling safe enough here to publish some not-so-great code.)

### Proposed changes

- introduce a USR2 signal to print all the Threads' backtraces
- generally keep the signals uppercased

### DIFF
--- a/lib/hutch/waiter.rb
+++ b/lib/hutch/waiter.rb
@@ -3,8 +3,12 @@ require 'hutch/logging'
 module Hutch
   class Waiter
     include Logging
-
+    
+    class SkipHere < Exception
+    end
+    
     SHUTDOWN_SIGNALS = %w(QUIT TERM INT).keep_if { |s| Signal.list.keys.include? s }.freeze
+    USER_SIGNALS = %w(USR1 USR2).keep_if { |s| Signal.list.keys.include?(s) }.freeze
 
     def self.wait_until_signaled
       new.wait_until_signaled
@@ -14,10 +18,20 @@ module Hutch
       self.sig_read, self.sig_write = IO.pipe
 
       register_signal_handlers
-      wait_for_signal
+      
+      begin
+        wait_for_signal
 
-      sig = sig_read.gets.strip.downcase
-      logger.info "caught sig#{sig}, stopping hutch..."
+        sig = sig_read.gets.strip.downcase
+        if USER_SIGNALS.include?(sig.upcase)
+          logger.info "FUN TIME! #{sig}"
+          raise SkipHere
+        else
+          logger.info "caught sig#{sig}, stopping hutch..."
+        end
+      rescue SkipHere
+        retry
+      end
     end
 
     private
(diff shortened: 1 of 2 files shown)


lib/hutch/waiter.rb
```diff
@@ -14,10 +18,20 @@ def wait_until_signaled
...
-
-      sig = sig_read.gets.strip.downcase
-      logger.info "caught sig#{sig}, stopping hutch..."
+      
+      begin
+        wait_for_signal
+
+        sig = sig_read.gets.strip.downcase
+        if USER_SIGNALS.include?(sig.upcase)
+          logger.info "FUN TIME! #{sig}"
```
michaelklishin: I appreciate the humour but perhaps we should make this log message a bit more descriptive :)

olleolleolle: Ah, I should have written "This is a handler for signal SIGUSR1, the behavior of which we will define as a community".

The current implementation is only a skeleton logging "proof" that it works.

michaelklishin:
Some common signals are `TERM` and `INT`, `HUP` (e.g. for config reloading — I don't think Hutch needs this, tbh), `USR1` and `USR2` (meaning entirely user defined).

If you don't have any specific ideas about what Hutch would do when sent a `USR1` or `USR2`,
I'm not sure we should trap them at all.

eebs:
It might be useful to allow pausing/resuming of the consumers. I see that Bunny could support pausing the work pool, but I didn't see anything in March Hare after a 2 minute glance.

michaelklishin:
The only way to pause a consumer is to stop deliveries to it by tweaking QoS prefetch. Otherwise you will run out of memory quickly. This sounds like a dangerous, hard to get right idea to me.

eebs:
Ah okay sorry, I was looking at Bunny's `ConsumerWorkPool#pause` and `Channel#maybe_pause_consumer_work_pool!`, but it looks like those aren't actually used anywhere.

olleolleolle:
I rebased.

michaelklishin:
So `USR2` performs a thread dump. I like this idea. Do we leave the door open for `USR1` handling for now then?

lib/hutch/waiter.rb
```diff
@@ -14,28 +25,70 @@ def wait_until_signaled
...
+      when 'USR2' then handle_usr2
+      else raise "Assertion failed - unhandled signal: #{sig.inspect}"
+      end
+      raise ContinueProcessingSignals
     end
 
     private
 
+    def handle_usr2
+      Thread.list.each do |thread|
```
michaelklishin: I suggest that we log a header such as "requested a VM-wide thread stack trace dump…"

lib/hutch/waiter.rb
```diff
@@ -14,28 +25,70 @@ def wait_until_signaled
...
+        logger.warn "Thread TID-#{thread.object_id.to_s(36)} #{thread['label']}"
```
michaelklishin: Why `warn` and not `info`?

lib/hutch/waiter.rb
olleolleolle: Agreed!

PR #264, opened 2016-11-19
Gemfile with ruby version marker, instead of specific gemfiles/ files

olleolleolle:
This PR changes the Gemfile for the project, to include the Bundler feature [`ruby RUBY_VERSION`](https://bundler.io/v1.13/man/gemfile.5.html#RUBY), so that Bundler can locate the right gems for the current Ruby version.

Current status: ~experimental~ This passes all Travis tests, and does the right thing.

- hutch.gemspec: Avoid concatenating a String by using line-continuation operator instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,12 @@ cache: bundler
 before_install: gem i bundler
 matrix:
   include:
-    - rvm: 2.3.0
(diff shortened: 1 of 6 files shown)


michaelklishin:
Thank you!

PR #265, opened 2016-11-24
Configure consumer tag names/prefixes

sldblog:
Allows us to configure the consumer names (tags) that are visible in RabbitMQ.

Tagging with specific names is useful for us because we have over 100 queues and lots of applications communicating. If we aim to be easily identifiable, we can do these things:

- define a routing key convention that refers to the application publishing the message
- define a queue name convention that identifies the consumers
- identify consumers by tag names

Out of the above three we cannot do the last one. So this changeset enables that configuration :)

### Notes

The default tag name changes from `bunny` to `hutch` with this change.

### Bonus

Explains on how to (still manually) update the generated configuration section in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,11 @@ In order from lowest to highest precedence:
 
 ### Generated list of configuration options
 
+Generate with
+
+0. `yard doc lib/hutch/config.rb`
+0. Copy the _Configuration_ section from `doc/Hutch/Config.html` here, with the anchor tags stripped.
+
 <table border="1" class="settings" style="overflow:visible;">
   <thead>
     <tr>
@@ -593,6 +598,15 @@ In order from lowest to highest precedence:
       <td>Boolean</td>
       <td><tt>HUTCH_CONSUMER_POOL_ABORT_ON_EXCEPTION</tt></td>
       <td><p>Should Bunny's consumer work pool threads abort on exception.</p>
+</td>
+    </tr>
+
+    <tr>
+      <td><tt>consumer_tag_prefix</tt></td>
+      <td>hutch</td>
+      <td>String</td>
+      <td><tt>HUTCH_CONSUMER_TAG_PREFIX</tt></td>
+      <td><p>Prefix displayed on the consumers tags.</p>
 </td>
     </tr>
   
--- a/lib/hutch/worker.rb
+++ b/lib/hutch/worker.rb
@@ -4,6 +4,7 @@ require 'hutch/broker'
 require 'hutch/acknowledgements/nack_on_all_failures'
 require 'hutch/waiter'
 require 'carrot-top'
+require 'securerandom'
 
 module Hutch
   class Worker
@@ -44,7 +45,7 @@ module Hutch
       queue = @broker.queue(consumer.get_queue_name, consumer.get_arguments)
       @broker.bind_queue(queue, consumer.routing_keys)
 
-      queue.subscribe(consumer_tag: consumer_tag(queue), manual_ack: true) do |*args|
+      queue.subscribe(consumer_tag: unique_consumer_tag, manual_ack: true) do |*args|
         delivery_info, properties, payload = Hutch::Adapter.decode_message(*args)
         handle_message(consumer, delivery_info, properties, payload)
       end
@@ -104,9 +105,10 @@ module Hutch
 
     attr_accessor :setup_procs
 
-    def consumer_tag(queue)
+    def unique_consumer_tag
       prefix = Hutch::Config[:consumer_tag_prefix]
-      queue.channel.generate_consumer_tag(prefix)
+      unique_part = SecureRandom.uuid
+      "#{prefix}-#{unique_part}"
     end
   end
 end
--- a/lib/hutch/worker.rb
+++ b/lib/hutch/worker.rb
@@ -44,7 +44,7 @@ module Hutch
       queue = @broker.queue(consumer.get_queue_name, consumer.get_arguments)
       @broker.bind_queue(queue, consumer.routing_keys)
 
-      queue.subscribe(manual_ack: true) do |*args|
+      queue.subscribe(consumer_tag: consumer_tag(queue), manual_ack: true) do |*args|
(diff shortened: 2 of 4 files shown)


michaelklishin:
@olleolleolle thoughts? This looks reasonable to me.

olleolleolle:
The feature is an improvement!

The failing test is about supporting MarchHare + Bunny: JRuby support.

```
LoadError: no such file to load -- bunny/channel
```

In order to get that working, perhaps we need to not use `instance_double()`?

lib/hutch/worker.rb
```diff
@@ -103,5 +103,10 @@ def error_acknowledgements
     private
 
     attr_accessor :setup_procs
+
+    def consumer_tag(queue)
+      prefix = Hutch::Config[:consumer_tag_prefix]
+      queue.channel.generate_consumer_tag(prefix)
```
olleolleolle: So, MarchHare does not have the same interface:

https://github.com/ruby-amqp/march_hare/search?utf8=%E2%9C%93&q=generate_consumer_tag

olleolleolle: Ah, it _does_ exist, as another way of expressing it: https://github.com/ruby-amqp/march_hare/pull/93/files

sldblog:
Oh I understand why `instance_double` wasn't used now. I didn't realise the underlying library is different on JRuby. Let me rework some of this.

README.md
```diff
@@ -334,6 +334,11 @@ In order from lowest to highest precedence:
 
 ### Generated list of configuration options
 
+Generate with
+
+0. `yard doc lib/hutch/config.rb`
+0. Copy the _Configuration_ section from `doc/Hutch/Config.html` here, with the anchor tags stripped.
```
olleolleolle: (This is you coming back to improve the README: 🎩 !)

spec/hutch/worker_spec.rb
```diff
@@ -45,9 +45,17 @@
     end
 
     it 'sets up a subscription' do
-      expect(queue).to receive(:subscribe).with(manual_ack: true)
+      expect(queue).to receive(:subscribe).with(consumer_tag: %r(^hutch\-.{36}$), manual_ack: true)
```
sldblog: Do I need to double check AMQP/RabbitMQ spec for the maximum available consumer tag name?

olleolleolle: That would be a cool touch.

(Then we could figure out ways to validate configuration options. Example addition to the Config options' DSL: `max_length: 34`.)

olleolleolle: It "depends", it seems: https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.consume.consumer-tag

sldblog: Seems like it's 255 bytes max: https://github.com/rabbitmq/rabbitmq-common/blob/rabbitmq_v3_6_0/codegen.py#L178-L179 ... I hope that's the same in 3.4.x as well :)

lib/hutch/worker.rb
```diff
@@ -103,5 +104,11 @@ def error_acknowledgements
     private
 
     attr_accessor :setup_procs
+
+    def unique_consumer_tag
+      prefix = Hutch::Config[:consumer_tag_prefix]
+      unique_part = SecureRandom.uuid
+      "#{prefix}-#{unique_part}"
```
sldblog: I ended up generating this myself. Quote from the commit:

> Due to the interface differences between `bunny` (used for CRuby/MRI)
and `march_hare` (used for JRuby), it's not possible to consistently use
the generator function `queue.channel.generate_consumer_tag` for the
unique part of the consumer tags.

> Instead we can generate the unique part ourselves without relying on
internals of these gems. Using `SecureRandom.uuid` seems like a good
idea.

Are these assumptions valid?

sldblog:
⌛️ I also need to test these with my local RabbitMQ install to ensure it's not breaking for some other integration problems.

spec/hutch/worker_spec.rb
olleolleolle: So, now we know that the longest prefix that can be used is: 256 - 1 - 36 = 219 bytes.

lib/hutch/worker.rb
michaelklishin: It is. `SecureRandom` is available on JRuby and produces values of a known fixed length (36 bytes).

spec/hutch/worker_spec.rb
michaelklishin: Ideally Hutch should loudly complain about prefixes that are longer, e.g. log an error and trim the prefix or fail to start.

sldblog: Opted for an "on-create" validation as validating the actual prefix length is weird (why 219 bytes?) and coupled to the implementation.

sldblog:
> I also need to test these with my local RabbitMQ install to ensure it's not breaking on some other integration problems.

💚 

#### Default (now)
screen shot 2016-11-25 at 15 27 21

#### Custom prefix
screen shot 2016-11-25 at 15 28 20

#### Oh dear
screen shot 2016-11-25 at 15 29 54

spec/hutch/worker_spec.rb
sldblog: Is the on-create validation sufficient/ok?

michaelklishin:
Consumer tags do not change and are rarely used in delivery handlers (probably never in Hutch).

sldblog:
> Consumer tags do not change and are rarely used in delivery handlers (probably never in Hutch).

@michaelklishin Does that mean it's okay to validate them when creating the subscriptions?

michaelklishin:
It does to me.

sldblog:
🎉 🙇

sldblog:
@michaelklishin Would it be possible to release this as 0.24? If it's blocked on something, can I help unblocking it?

michaelklishin:
Yes.

michaelklishin:
`0.24.0` is out.

sldblog:
❤️

PR #270, opened 2016-12-08
Expose uri as normal config setting, so it can be used in ENV

### LINKED ISSUE #269: HUTCH_URI environment variable doesn't work
Hi,

I see uri is available via Hutch::Config but it's not available as environment variable.. is this intentional? I'm considering writing a PR for this functionality.

samstickland:
Note, I see in the logs that this shouldn't be used from the CLI, but I see that the CLI class is manually processing it's options so this work happen as a result of this.

Also updates the readme to show the URI option (via yard generation).

Fixes #269.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,6 +431,15 @@ Generate with
     </tr>
   
     <tr>
+      <td><tt>uri</tt></td>
+      <td>nil</td>
+      <td>String</td>
+      <td><tt>HUTCH_URI</tt></td>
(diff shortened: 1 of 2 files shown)


michaelklishin:
@olleolleolle WDYT?

olleolleolle:
Got me thinking - should the yard outputter sort its list of ENV vars on their name?

PR #271, opened 2017-01-20
Travis: Add 2.4.0, use jruby-9.1.7.0; don't trap QUIT on JRuby since it's not supported

olleolleolle:
This PR

- adds 2.4.0 to the build matrix
- updates to latest JRuby

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,12 @@ cache: bundler
 before_install: gem i bundler
 matrix:
   include:
+    - rvm: 2.4.0
     - rvm: 2.3.3
     - rvm: 2.2
     - rvm: 2.1
     - rvm: 2.0
-    - rvm: jruby-9.1.6.0
+    - rvm: jruby-9.1.7.0
       jdk: oraclejdk8
       env:
         - JRUBY_OPTS='--debug'
--- a/lib/hutch/waiter.rb
+++ b/lib/hutch/waiter.rb
@@ -12,7 +12,9 @@ module Hutch
     end
 
     def self.supported_signals_of(list)
-      list.keep_if { |s| Signal.list.keys.include? s }
+      list.keep_if { |s| Signal.list.keys.include?(s) }.tap do |result|
+        result.delete('QUIT') if defined?(JRUBY_VERSION)
+      end
     end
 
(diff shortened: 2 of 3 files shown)


michaelklishin:
Looks like tests hang on JRuby now, presumably because `SIGQUIT` is not handled as it used to be.

olleolleolle:
@michaelklishin That message about SIGQUIT has been in JRuby a long time. Should Hutch change its tune about that signal?

michaelklishin:
@olleolleolle on JRuby only, I think so.

olleolleolle:
I was able to get the hang on QUIT with "Randomized with seed 60614".

michaelklishin:
@olleolleolle 👍  let's drop `QUIT` support on JRuby then.

michaelklishin:
By the way, we can do it in this very PR.

olleolleolle:
@michaelklishin I'll see if I can get a fix on the thing, so that we don't... mess things up.

olleolleolle:
@michaelklishin I began by unwinding a loop, so I could be remove a weirdly-placed `next`.

michaelklishin:
@olleolleolle I'm not sure what you mean. But thank you.

olleolleolle:
@michaelklishin The next move would be to never mention the QUIT signal to JRuby in the implementation of Waiter, so that no warnings are spouted.

michaelklishin:
Makes sense, thank you!

olleolleolle:
Now, the test output of WaiterSpec is

```
Hutch::Waiter
  ["TERM", "INT"]
    includes only things in Signal.list.keys
  .wait_until_signaled
    a TERM signal is received
      logs that hutch is stopping
    a INT signal is received
      logs that hutch is stopping
```

olleolleolle:
@michaelklishin Can you kill this build? https://travis-ci.org/gocardless/hutch/jobs/193808578

olleolleolle:
@michaelklishin I'm more interested in the _next_ build, which included the "skip QUIT on JRuby".

michaelklishin:
@olleolleolle thank you!

olleolleolle:
@michaelklishin Happy weekend!

PR #272, opened 2017-01-23
Travis: Rubygems update in before_install

olleolleolle:
While investigating if it would be possible to remove the `ruby RUBY_VERSION` marker in the Gemfile, I noted that we don't upgrade Rubygems before using it.

Do we want to?


michaelklishin:
Thank you!

PR #274, opened 2017-02-21
Wrap API of error handler for Honeybadger

olleolleolle:
The API method used is [removed in the upstream 3.0.0](https://github.com/honeybadger-io/honeybadger-ruby/blob/master/CHANGELOG.md#300---2017-02-06), which was just released (two, three weeks ago).

I read [the new method's docs](http://www.rubydoc.info/gems/honeybadger/Honeybadger%3Anotify) again, and noted that if the first arg is a Hash, then it's just 1 argument.


michaelklishin:
Thank you!

PR #281, opened 2017-05-17
Make hutch can use distinguish is run use `bin/hutch` ?

wppurking:
Removed `hutch/cli` require in lib/hutch.rb because it`s only used by `bin/hutch` to launched use command


michaelklishin:
May I ask what problem we are trying to solve here?

lib/hutch.rb
```diff
@@ -18,6 +17,10 @@ def self.register_consumer(consumer)
     self.consumers << consumer
   end
 
+  def self.server?
```
michaelklishin: You can run Hutch just for publishing and still use it in a server of some kind. A more suitable name would be `daemon?` or `consumer_service?` (mind you, neither is perfect).

spec/hutch_spec.rb
```diff
@@ -83,5 +83,11 @@
       Hutch.publish(*args)
     end
   end
+
+  # only works on `rspec spec/hutch_spec.rb` will be green.
+  # i don`t know why the Hutch::CLI is loaded when `rspec spec`
```
michaelklishin: Perhaps because [there are CLI tests](https://github.com/gocardless/hutch/blob/master/spec/hutch/cli_spec.rb).

michaelklishin:
I 👎 this. I'm not sure what problem it solves and making the new tests pass will be a challenge because `rspec` will load `hutch/cli` to test, ahem, `Hutch::CLI`. Is it really an improvement worth that additional complexity?

wppurking:
@michaelklishin Thanks for you quick response 😄 

> "May I ask what problem we are trying to solve here?"

When i work with rails and hutch, there is two process running.
- Rails run it`s own process to server web request and can use the Hutch.publish to send message.
- Hutch run it`s own process and load all the Rails code into memery to consum queue message.

The first thought is i can distinguish which process environment the code is running?
If Hutch::CLI is exist? so the process is running through Hutch command.
If Hutch::CLI is not exist? so the process is running in Rails process.

So if i can distinguish the two process environment, i can put some code is only need in Hutch process. Such as i just need use [`ApplicationJob.descendants`](http://api.rubyonrails.org/classes/ActiveSupport/DescendantsTracker.html) to get all ActiveJob class and pick outs all correct `queue_name`, `routing_key` mapped them to correct consumer with the same `#process` method.  If i already done this, the Hutch integrate with ActiveJob will be perfect solved.
(one queue and one consumer in hutch is mapping with one queue and one JobClass in ActiveJob,  right now is each queue with an JobClass in ActiveJob but mapping them to one queue and one consumer in Hutch)

Right now i am try to make this code into an `Hutch.setup_procs` proc.

spec/hutch_spec.rb
wppurking: I see this file, if this file is effect this that means when i run `rspec spec` it`s load all *_spec.rb into memery at once so the 'hutch/cli' is loaded before the `spec/hutch_spec.rb` run.

lib/hutch.rb
wppurking: Yes, maybe the `server?` is not the good name, i just borrow it from sidekiq 😆

wppurking:
@michaelklishin i found an other way to integrate with ActiveJob. I closed this pull request~ 
Thanks~

PR #282, opened 2017-05-24
when boot rails project, log rails environment

wppurking:
Add rails environment log information to Hutch::CLI.

When use `bundle exec hutch -v` to run hutch there is no exactly information about what is the rails environment is use.


michaelklishin:
Makes sense, thank you!

PR #290, opened 2017-09-15
Honey badger notify no force

biinari:
#274 wraps the Honeybadger API to use `notify` after the removal of `notify_or_ignore`. By default, `notify` behaves the same, ignoring notices based on configuration. With the `force` option set to `true`, this is changed to notify regardless of whether the notice was configured to be ignored.

Hence, I'm removing the `force` option to restore previous behaviour.


michaelklishin:
Looks reasonable. @olleolleolle let us know if there was a specific reason why `force: true` was introduced in #274.

olleolleolle:
@michaelklishin Just me misreading instructions! Carry on!

biinari:
Cool, thanks @michaelklishin and @olleolleolle

PR #291, opened 2017-09-18
Travis: rvm nowadays self-updates "gem"

olleolleolle:
This tiny PR removes a step which `rvm` now performs on its on.


michaelklishin:
Thank you!

PR #296, opened 2017-10-20
Group consumers

nickolai-sm:
In case you use hutch with rails and you no need to run all consumers. Or if you need to split then into groups to balance load. Then define groups in config file:

``` yml
consumers_groups:
  payments:
    - DepositConsumer
    - CashoutConsumer
  notification:
    - EmailNotificationConsumer
```

and run ``` hutch --only-group=payments --config=CONFIG_PATH```

### DIFF
--- a/lib/hutch/worker.rb
+++ b/lib/hutch/worker.rb
@@ -36,12 +36,18 @@ module Hutch
     # Set up the queues for each of the worker's consumers.
     def setup_queues
       logger.info 'setting up queues'
-      @consumers.each { |consumer| setup_queue(consumer) }
+      @consumers.each do |consumer|
+        next if group_configured? && group_restrict?(consumer)
+
+        setup_queue(consumer)
+      end
     end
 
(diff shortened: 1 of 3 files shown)


lib/hutch/worker.rb
```diff
@@ -36,12 +36,18 @@ def stop
     # Set up the queues for each of the worker's consumers.
     def setup_queues
       logger.info 'setting up queues'
-      @consumers.each { |consumer| setup_queue(consumer) }
+      @consumers.each do |consumer|
+        next if group_configured? && group_restrict?(consumer)
```
michaelklishin: if we use `group_configured?` then this should be `group_restricted?`

michaelklishin:
Should `--group` be `--only-group`?

nickolai-sm:
Yes, it will be clear

michaelklishin:
@nickolai-sm note that I took the liberty to slightly rename the config setting.

Thank you!

PR #301, opened 2017-12-05
Handle setup exceptions

### LINKED ISSUE #288: Connection error not reported to sentry

valentin-krasontovitsch:
To handle setup exceptions, a method was added to each error handler.

Said method is subsequently called when setup of the client fails.

This is to ensure that a failed setup is not only logged, but also
reported to e.g. sentry.

Closes #288


michaelklishin:
@sldblog OK with you to merge this?

PR #302, opened 2017-12-05
Add error handler class

### LINKED ISSUE #288: Connection error not reported to sentry

valentin-krasontovitsch:
To handle setup exceptions, a method was added to each error handler.

Said method is subsequently called when setup of the client fails.

This is to ensure that a failed setup is not only logged, but also
reported to e.g. sentry.

Furthermore, a base class indicating the interface for error handlers is added.

Closes #288

### DIFF
--- a/lib/hutch/cli.rb
+++ b/lib/hutch/cli.rb
@@ -90,6 +90,9 @@ module Hutch
       @worker.run
       :success
     rescue ConnectionError, AuthenticationError, WorkerSetupError => ex
+      Hutch::Config[:error_handlers].each do |backend|
+        backend.handle_setup_exception(ex)
+      end
       logger.fatal ex.message
       :error
     end
--- a/lib/hutch/error_handlers/base.rb
+++ b/lib/hutch/error_handlers/base.rb
@@ -1,0 +1,15 @@
+module Hutch
+  module ErrorHandlers
+    class Base
+      include Logging
+
+      def handle(properties, payload, consumer, ex)
+        raise NotImplementedError.new
+      end
+
+      def handle_setup_exception(ex)
+        raise NotImplementedError.new
+      end
+    end
+  end
+end
--- a/lib/hutch/error_handlers/airbrake.rb
+++ b/lib/hutch/error_handlers/airbrake.rb
@@ -1,10 +1,10 @@
 require 'hutch/logging'
 require 'airbrake'
+require 'hutch/error_handlers/base'
 
 module Hutch
   module ErrorHandlers
-    class Airbrake
-      include Logging
+    class Airbrake < Base
 
       def handle(properties, payload, consumer, ex)
         message_id = properties.message_id
--- a/lib/hutch/error_handlers/airbrake.rb
+++ b/lib/hutch/error_handlers/airbrake.rb
@@ -31,6 +31,24 @@ module Hutch
           })
         end
       end
+
+      def handle_setup_exception(ex)
+        logger.error "Logging setup exception to Airbrake"
+        logger.error "#{ex.class} - #{ex.message}"
+
+        if ::Airbrake.respond_to?(:notify_or_ignore)
+          ::Airbrake.notify_or_ignore(ex, {
+            error_class: ex.class.name,
+            error_message: "#{ ex.class.name }: #{ ex.message }",
+            backtrace: ex.backtrace,
+            cgi_data: ENV.to_hash,
+          })
+        else
+          ::Airbrake.notify(ex, {
+            cgi_data: ENV.to_hash,
+          })
+        end
+      end
     end
   end
 end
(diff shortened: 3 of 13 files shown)


lib/hutch/error_handlers/base.rb
```diff
@@ -0,0 +1,15 @@
+module Hutch
+  module ErrorHandlers
+    class Base
+      include Logging
+
+      def handle(properties, payload, consumer, ex)
+        raise NotImplementedError.new
```
sldblog: Unfortunately, `NotImplementedError` is subclassed from `ScriptError` which is subclassed from `Exception`.

I think it would be safer to use something that can be caught by `StandardError`.

valentin-krasontovitsch: There seems to be no standard way of doing this besides the above, and not throwing an Error, but relying on the no method error being thrown when a subclass doesn't implement the method.

Do you suggest to define a custom Error class that inherits from StandardError?

Please cf. #297 for links to discussions regarding the above error... I'm a newbie to ruby (heh, that rhymed), so I really don't feel confident to make a call, due to the different opinions I found...

Also, we should probably fix [the ack base class](https://github.com/gocardless/hutch/blob/fa9383a40cfc4e681c4128e3cc47e84c8820546f/lib/hutch/acknowledgements/base.rb#L12) as well if we mind the above.

michaelklishin: `NotImplementedError` is right on the money in terms of what we'd like to communicate, though.

valentin-krasontovitsch: Again, I don't really know, not a seasoned ruby programmer, but: the docs for the `NotImplementedError` can be interpreted in such a way that it is only to be used for instanced where a feature is not implemented on the current platform, due to a hardware / software deficit of the underlying system, not a missing implementation in the ruby code.
Even the error message in `error.c`, where the error class is defined, sounds as follow:
`[$FUNCTION_NAME() function is unimplemented on this machine](https://github.com/ruby/ruby/blob/1ae69a20645d7be556fd0ecd001b0bb5f3959330/error.c#L2338)`

Of course I don't actually get that message when I raise the error in irb, just get the text `NotImplementedError`...

The above are just meant as hints - you guys should decide how to handle this problem : )

sldblog: I don't think this discussion has to delay this PR, it's great :)

I have only called this out because in April this year we had apps not working due to a missed implementation, and we did not get any Honeybadger reports since it as `NotImplementedError` and zipped past through any layers of error reporting so we didn't know what was happening before looking at the logs.

lib/hutch/cli.rb
```diff
@@ -90,6 +90,9 @@ def start_work_loop
       @worker.run
       :success
     rescue ConnectionError, AuthenticationError, WorkerSetupError => ex
+      Hutch::Config[:error_handlers].each do |backend|
+        backend.handle_setup_exception(ex)
+      end
```
sldblog: One thing though, could this be tested in `cli_spec`, please?

lib/hutch/error_handlers/base.rb
michaelklishin: @sldblog the fact that in Ruby a `rescue` without specific exceptions doesn't catch everything not a problem library authors should worry about. Ruby designers made a mistake,  using a `rescue` without an exception class is therefore is not a good idea. `NotImplementedError` is exactly the exception that should be used in this case.

valentin-krasontovitsch:
I can’t say that I didn’t look at adding a test, but I was a wee bit
stumped as to what it might look like. You got me ^^

I’ll take another look and try adding a test : )

michaelklishin:
@valentin-krasontovitsch thank you.

valentin-krasontovitsch:
So I just pushed a commit that introduces a test case, making sure that if `Hutch.connect` throws an error, that error is passed to all backends.

I'm not super happy with the test, as it relies on the implementation of the tested method: `start_work_loop` is tested, and we assume that the method calls `Hutch.connect`, which in turn is used to raise an exception.

If you have any pointers or ideas about how this could be tested in a less brittle way, let me know.

Maybe the coupling to the implementation is OK in this case...

valentin-krasontovitsch:
So I'm basically happy with the other merge request merged in, but just for completeness sake: Any new thought / comments on the test I wrote?

michaelklishin:
@valentin-krasontovitsch sorry, this slipped through the cracks. What's the other PR you are referring to? Is this one still relevant and would you like us to review it?

valentin-krasontovitsch:
no worries : )
the other PR was #301, included the first two commits of this merge request, and got merged immediately - while here (rightfully) the lacking unit tets was pointed out (and mitigated).

I guess as you've merged this one now, i don't need to answer your second question anymore ^^

PR #305, opened 2018-01-16
When using the uri config option: Respect amqps specification and use default protocol ports if unspecified

michaellennox:
### What does this PR do?

Restores functionality implemented in https://github.com/gocardless/hutch/pull/159 and since lost to:

* Respect `amqps` specification when using the URI config option (ie. utilise TLS)
* Default to the standard protocol ports when not specified and using the URI config option (5671 for TLS, 5672 otherwise)

This means that:

`amqp://guest:guest@127.0.0.1/` connects to the server on port 5672 and does not utilise TLS.
`amqps://guest:guest@127.0.0.1/` connects to the server on port 5671 and does utilise TLS.

It also adds test cases for this functionality to prevent another regression.

### Where should the reviewer start?

Changes are all within `Hutch::Broker#parse_uri`

### Any background context you want to provide?

Background context can be found in the original issue ://github.com/gocardless/hutch/issues/304

### Any specific implementation details?

The test for covering when amqps scheme is utilised has to stub out `Hutch::Adapter` instead of acting like an integration test in a similar manner to the surrounding specs because it appears a default rabbitmq install does not activate TLS listeners and as such a user who was to run the test suite without changing their rabbitmq server config would get failing specs if the adapter was not stubbed.

\cc @michaelklishin

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -292,11 +292,16 @@ module Hutch
 
       u = URI.parse(@config[:uri])
 
+      @config[:mq_tls]      = u.scheme == 'amqps'
       @config[:mq_host]     = u.host
-      @config[:mq_port]     = u.port
+      @config[:mq_port]     = u.port || default_mq_port
       @config[:mq_vhost]    = u.path.sub(/^\//, "")
       @config[:mq_username] = u.user
       @config[:mq_password] = u.password
+    end
+
+    def default_mq_port
+      @config[:mq_tls] ? AMQ::Protocol::TLS_PORT : AMQ::Protocol::DEFAULT_PORT
     end
 
     def sanitized_uri
--- a/spec/hutch/broker_spec.rb
+++ b/spec/hutch/broker_spec.rb
@@ -92,6 +92,36 @@ describe Hutch::Broker do
 
       connection.close
     end
+
+    context 'when configured with a URI' do
+      context 'which specifies the port' do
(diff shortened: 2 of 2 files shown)


lib/hutch/broker.rb
```diff
@@ -292,13 +292,18 @@ def parse_uri
...
       @config[:mq_host]     = u.host
-      @config[:mq_port]     = u.port
+      @config[:mq_port]     = u.port || default_mq_port
       @config[:mq_vhost]    = u.path.sub(/^\//, "")
       @config[:mq_username] = u.user
       @config[:mq_password] = u.password
     end
 
+    def default_mq_port
+      @config[:mq_tls] ? 5671 : 5672
```
michaelklishin: Please switch to `AMQ::Protocol::DEFAULT_PORT` and `AMQ::Protocol::TLS_PORT`.

michaellennox: Thanks @michaelklishin , ♻️

michaellennox:
Are failures in the JRuby build expected?

Looking at the travis output all the specs passed but travis marked the build as a failure as it believed the build timed out: https://travis-ci.org/gocardless/hutch/jobs/329386746

michaelklishin:
With JRuby you have to properly close connections or the JVM won't ever terminate because RabbitMQ Java client will have an execution service with running non-daemon threads. So it's not related.

PR #306, opened 2018-01-18
Mention GoCardless origin but remove "join us" footer

nickcampbell18:
This small change is to emphasise that whilst the majority of the work was completed at GC, it's now a community project since we don't actively use it at GC 🎉


nickcampbell18:
Thanks @michaelklishin !

PR #313, opened 2018-05-29
Drop now-defunct Opbeat error handler and tracer

olleolleolle:
This PR removes the **Opbeat** error handler and tracer.

  - the Opbeat service is no longer generally available

Hugs to everyone who ever loved Opbeat.  :hearts: :yellow_heart: :green_heart: :blue_heart: :purple_heart:


michaelklishin:
Thank you!

PR #315, opened 2018-08-16
Fix NameError: undefined local variable or method `allowed_consumers`

nickolai-sm:
It seems during rename(https://github.com/gocardless/hutch/commit/d509c19763902e6ebf07a57d78a24431f254015c#diff-16eb6365e2d1f19f4df2b8aba16e4f9cL121) took place a mistake.


michaelklishin:
Thank you!

PR #332, opened 2019-07-02
Added error handler for Rollbar

johankok:
This Pull Request adds an error handler that reports exceptions to Rollbar.


michaelklishin:
Thank you!

PR #337, opened 2019-11-08
Call confirm_select at publishing time

### LINKED ISSUE #189: Prevent some configuration changes after running `Hutch.connect`
Today I spent some time debugging an issue that was caused by me doing this:

``` ruby
Hutch.connect
# ... 
Hutch::Config.set(:force_publisher_confirms, true)
Hutch.publish(blah)
```

This doesn't work because `.confirm_select` has to be called on the channel before calling `.wait_for_confirms`.

`.confirm_select` is only called in `Hutch.connect` if `force_publisher_confirms` is set. So setting this _after_ running `Hutch.connect` will cause `Hutch.publish` to fail with a nil error:

``` ruby
NoMethodError: undefined method `empty?' for nil:NilClass
    from /Users/jesjos/.rvm/gems/ruby-2.3.0/gems/bunny-2.2.2/lib/bunny/channel.rb:1812:in `block in wait_on_confirms_continuations'
    from /Users/jesjos/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/monitor.rb:214:in `mon_synchronize'
    from /Users/jesjos/.rvm/gems/ruby-2.3.0/gems/bunny-2.2.2/lib/bunny/channel.rb:1811:in `wait_on_confirms_continuations'
    from /Users/jesjos/.rvm/gems/ruby-2.3.0/gems/bunny-2.2.2/lib/bunny/channel.rb:1434:in `wait_for_confirms'
    from /Users/jesjos/.rvm/gems/ruby-2.3.0/gems/hutch-0.20.0/lib/hutch/broker.rb:227:in `publish'
    from /Users/jesjos/.rvm/gems/ruby-2.3.0/gems/hutch-0.20.0/lib/hutch.rb:60:in `publish'
```

I'm thinking it would be nice if Hutch somehow failed early or at least warned about this. Maybe it would make sense to prevent some, or all, kinds of configuration changes once Hutch is connected.

I'd gladly provide a PR if and once an appropriate strategy is identified.

arthurmde:
Fix #189

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -109,10 +109,6 @@ module Hutch
       logger.info "opening rabbitmq channel with pool size #{consumer_pool_size}, abort on exception #{consumer_pool_abort_on_exception}"
       connection.create_channel(nil, consumer_pool_size, consumer_pool_abort_on_exception).tap do |ch|
         connection.prefetch_channel(ch, @config[:channel_prefetch])
-        if @config[:publisher_confirms] || @config[:force_publisher_confirms]
-          logger.info 'enabling publisher confirms'
-          ch.confirm_select
-        end
       end
     end
 
--- a/lib/hutch/publisher.rb
+++ b/lib/hutch/publisher.rb
@@ -19,6 +19,11 @@ module Hutch
 
       serializer = options[:serializer] || config[:serializer]
 
+      if @config[:publisher_confirms] || @config[:force_publisher_confirms]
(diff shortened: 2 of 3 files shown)


michaelklishin:
Thank you but I'm afraid I cannot accept this solution. Doing this on the hot code path makes little sense — confirms have to be enabled only once — and will introduce the amount of overhead that you likely underestimate. Enabling confirms is a protocol method (operation), so for *every message published* this would send an additional frame for no real reason.

PR #341, opened 2020-01-30
Add lazy and quorum options for queues

### LINKED ISSUE #340: Enable Lazy mode for queues
It would be interesting if Hutch offered options to set the consumers' queues as [lazy](https://www.rabbitmq.com/lazy-queues.html) through an option such as `queue_name`.

[Lazy queues are recommended in some use cases.](https://www.cloudamqp.com/blog/2017-12-29-part1-rabbitmq-best-practice.html#enable-lazy-queues-to-get-predictable-performance)

arthurmde:
Fix #340

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,28 @@ class FailedPaymentConsumer
 end
 ```
 
+It is possible to set some custom options to consumer's queue explicitly.
+This example sets the consumer's queue as a
+[quorum queue](https://www.rabbitmq.com/quorum-queues.html)
+and to operate in the [lazy mode](https://www.rabbitmq.com/lazy-queues.html).
+The `initial_group_size`
+[argument](https://www.rabbitmq.com/quorum-queues.html#replication-factor) is
+optional.
+
+```ruby
+class FailedPaymentConsumer
+  include Hutch::Consumer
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -109,6 +109,26 @@ class FailedPaymentConsumer
...
+and to operate in the [lazy mode](https://www.rabbitmq.com/lazy-queues.html).
+The `initial_group_size` argument is optional.
+
+```ruby
+class FailedPaymentConsumer
+  include Hutch::Consumer
+  consume 'gc.ps.payment.failed'
+  queue_name 'failed_payments'
+  lazy_queue
+  quorum_queue initial_group_size: 5
```
michaelklishin: Can we please  use  a  lower value? Examples like this tend to be copied straight into production and plastered all over tutorials on the Web, and I'm not sure if everyone wants to run with at least 5 replicas for every queue :)

arthurmde: Sure, I will use 3 which is the minimum value that [makes practical sense](https://www.rabbitmq.com/quorum-queues.html#replication-factor)

michaelklishin:
Thank you!

PR #343, opened 2020-02-19
Better log levels in the message publisher

sdwolfz:
Having the entire payload logged at the "INFO" level is not ideal due to
the following reasons:
- PII data might be contained there and there is no way to fitler it out
- Payload might be big enough that logging aggregators might struggle with
the sheer amount of data
- Since the log level is shared by reusing the rails/sinatra logger, INFO level
might be too verbose for hutch but WARN level not verbose enough for the other.

While initially I considered implementing a feature to allow for filtering
the payload for PII or triming the output, a quick solution to all problems
listed above is to split the logging into two, the INFO level handing the
generic part of the log, with the message ID injected for easy tracking
and in case the payload is needed, log level can be set to DEBUG, which
should only be needed in development/testing environments.

### DIFF
--- a/lib/hutch/broker.rb
+++ b/lib/hutch/broker.rb
@@ -184,7 +184,7 @@ module Hutch
 
       filtered = api_client.bindings.
         reject { |b| b['destination'] == b['routing_key'] }.
-        filter { |b| b['source'] == @config[:mq_exchange] && b['vhost'] == @config[:mq_vhost] }
+        select { |b| b['source'] == @config[:mq_exchange] && b['vhost'] == @config[:mq_vhost] }
 
       filtered.each do |binding|
         results[binding['destination']] << binding['routing_key']
@@ -197,7 +197,7 @@ module Hutch
     def unbind_redundant_bindings(queue, routing_keys)
       return unless http_api_use_enabled?
 
-      filtered = bindings.filter { |dest, keys| dest == queue.name }
+      filtered = bindings.select { |dest, keys| dest == queue.name }
       filtered.each do |dest, keys|
         keys.reject { |key| routing_keys.include?(key) }.each do |key|
           logger.debug "removing redundant binding #{queue.name} <--> #{key}"
--- a/lib/hutch/publisher.rb
+++ b/lib/hutch/publisher.rb
@@ -42,7 +42,7 @@ module Hutch
     private
 
     def log_publication(serializer, payload, routing_key)
-      logger.info {
+      logger.debug {
         spec =
           if serializer.binary?
             "#{payload.bytesize} bytes message"


lib/hutch/publisher.rb
```diff
@@ -27,30 +27,29 @@ def publish(routing_key, message, properties = {}, options = {})
...
-        spec =
-          if serializer.binary?
-            "#{payload.bytesize} bytes message"
-          else
-            "message '#{payload}'"
-          end
-        "publishing #{spec} to #{routing_key}"
-      }
+    def log_publication(serializer, payload, payload_properties, routing_key)
+      logger.info("publishing message with id: #{payload_properties[:message_id]} to #{routing_key}")
```
sldblog: _🙋‍♂generic observation_

It seems `routing_key` is included in `payload_properties` due to the `non_overridable_properties` above.

What do you think about using `def log_publication(serializer, payload, message_id, routing_key)` as the signature to be more specific and avoid passing down the `routing_key` twice?

sdwolfz: Sounds good @sldblog!

sdwolfz:
I pushed a commit to fix the tests on older ruby versions; filter is an alias of select. Please review them independently.

lib/hutch/publisher.rb
```diff
@@ -41,16 +41,15 @@ def publish(routing_key, message, properties = {}, options = {})
...
-      }
+    def log_publication(serializer, payload, routing_key, message_id)
+      logger.info("publishing message with id: #{message_id} to #{routing_key}")
+
+      spec = if serializer.binary?
+        "publishing a #{payload.bytesize} bytes message"
+      else
+        "publishing message: '#{payload}'"
+      end
+      logger.debug(spec)
```
sldblog: 💭 It looks like the original code used the block-variant of `logger.info`. According to [the `Logger` docs](https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger.html#class-Logger-label-How+to+log+a+message), that has lazy evaluation.

Would it be worth to use the block-type `logger.debug { ... }` to avoid evaluating the `spec =` on every log level?

lib/hutch/broker.rb
```diff
@@ -184,7 +184,7 @@ def bindings
 
       filtered = api_client.bindings.
         reject { |b| b['destination'] == b['routing_key'] }.
-        filter { |b| b['source'] == @config[:mq_exchange] && b['vhost'] == @config[:mq_vhost] }
+        select { |b| b['source'] == @config[:mq_exchange] && b['vhost'] == @config[:mq_vhost] }
```
michaelklishin: How is this change relevant?

sdwolfz: This fixes build failures for older ruby versions, see travis builds: https://travis-ci.org/gocardless/hutch/jobs/652451121?utm_medium=notification&utm_source=github_status

michaelklishin: Understood, thanks.

sdwolfz:
> Actually, should we log publishes at debug level? Publishes can happen at a rate of hundreds or thousands per second, I really doubt it makes much sense to log at info level regardless of message size.

I'm fine with any solution as long as the current verbosity goes down. Just let me know what changes you want me to make.

michaelklishin:
@sdwolfz let's keep the changes you have but change publish logging level to `debug`. Thank you.

sdwolfz:
> @sdwolfz let's keep the changes you have but change publish logging level to `debug`. Thank you.

Are you sure you want to have two `logging.debug` calls there? I could revert the entire change and not pass the `message_id` anymore, just having the `info` call replaced with `debug`.

michaelklishin:
Good point. I think simply switching to debug logging would be best.

sdwolfz:
@michaelklishin all done! ♻️ 🙏

michaelklishin:
Thank you!

sdwolfz:
Cheers mate! 

Any hint on when this will be released to rubygems?

michaelklishin:
We haven't updated release notes in a while so it will take some time. Hopefully in the next few days.

sdwolfz:
@michaelklishin Can I help with that? I assume I need to open a PR with changelog lines for all PRs since the last tag.

PR #345, opened 2020-03-12
Update README.md change autoloads

lazing:
Dir.glob(File.join('**', '*_consumer.rb')) will load gem bunny guide class, result in rails startup failure.


michaelklishin:
Thank you!

lazing:
> Thank you!

Thanks for the great gem

PR #356, opened 2020-08-06
Add namespace to documentation + Enforce `MultiJson` usage

BuonOmo:
Hi,

I've made a PR for two distinct reasons (hence the two commits!)

**Add namespace to the readme** 08c8eaf
It seems like a pretty valid config, let's expose it!

**enforce MultiJson usage** 6cfb105
This allows users to always have their chosen implementation used. And gives a more consistent behavior.

Thanks for the great tool 🚀


michaelklishin:
Thank you!

PR #357, opened 2021-01-04
Correct typo in README

nikokon:
Seems like there was a typo regarding the key under witch consumer groups were configured.


michaelklishin (approved):

michaelklishin:
Good catch, thank you!

PR #358, opened 2021-01-15
handle type-casting for config, even when explicitly setting values using .set method

Azdaroth:
I had a pretty unfortunate issue when directly setting the config value for `heartbeat` without an explicit typecasting: `Hutch::Config.set(:heartbeat, ENV.fetch("HUTCH_HEARTBEAT", 30))` but the ENV var was obviously a string and it caused some failures when publishing messages. Since the type is explicitly declared on a config level, I believe the desired behaviour would be to handle type-casting inside `set` method as well.

### DIFF
--- a/lib/hutch/config.rb
+++ b/lib/hutch/config.rb
@@ -196,14 +196,7 @@ module Hutch
       env_keys_configured.each_with_object({}) {|attr, result|
         value = ENV[key_for(attr)]
 
-        case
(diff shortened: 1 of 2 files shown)


Azdaroth:
Hey, any news regarding this PR? Does it look like the desired behavior?

michaelklishin:
This looks reasonable, I need to take a look at the test failure.

michaelklishin (approved):

PR #360, opened 2021-03-02
add Datadog tracer

Azdaroth:
There is already a tracer for NewRelic. Having a tracer for Datadog as well sounds like a good idea then.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #363, opened 2021-04-06
add 'literal' Sentry error handler and make it separate from legacy Raven

Azdaroth:
Turns out that Sentry as Raven is deprecated and is going to be replaced by literal Sentry, coming from `sentry-ruby` SDK. So I renamed the existing error handler to SentryRaven and added a new Sentry one.


michaelklishin (approved):

michaelklishin:
The only failure is on Ruby 2.3. Let's test on 2.4+, it is four years old already.

PR #372, opened 2021-11-11
Make Bunny all queue options available

kavv-hub:
want to make use of bunny queue options, e.g. disallow application to create queue by setting `no_declare: true`


michaelklishin:
This does not seem to be a breaking change, so adopt this idea. Thank you.

PR #376, opened 2022-06-03
Fix $LOAD_PATH error

AlexanderMint:
Fixes a bug with requiring components.

Example: https://github.com/AlexanderMint/hutch_example_error

If an executable file has paths added to $LOAD_PATH, then the last path added by the user will be removed


michaelklishin:
Thank you. This makes sense to me.

PR #386, opened 2022-09-02
fix Sentry integration - make sure that the payload is a hash, not JSON

Azdaroth:
I've just discovered in one of the projects that errors are not getting delivered to Sentry and in the logs I found something coming from `lib/sentry/utils/argument_checking_helper`: `ArgumentError: expect the argument to be a Hash, got String` 

Looks like the setup was not done correctly in the spec and `configure_scope` method was returning early, without the evaluation of the block (https://github.com/getsentry/sentry-ruby/blob/c9b4eac913f0c6510178c451da57d91fc340fa20/sentry-ruby/lib/sentry-ruby.rb#L163). And because of that, the spec didn't catch the error related to the expected type of `payload` which must be a hash, not a JSON. This patch fixes it.


Azdaroth:
Hmm, not sure why the suite is failung under JRuby, looks unrelated :/

michaelklishin:
Yes, the JRuby failures are not related or new.

PR #390, opened 2022-12-14
Allow auth mechanism to be configured

ball-hayden:
Exposes the ability to configure Bunny's `auth_mechanism` parameter, so that TLS certificate authentication may be used.

See http://rubybunny.info/articles/connecting.html for details.


ball-hayden:
@michaelklishin thanks for approving the CI run.

I'm confused by the failures - do you know if any of those are regular flakes?

michaelklishin:
They are not really flakes, JRuby CI has a history of failures for Hutch.

PR #393, opened 2023-01-09
exact queue name option

tlloydthwaites:
Sometimes we want to bind to an exact queue name without using the namespace.

For example, for most of our consumers we want to use the application name as a prefix to help group queues. However, we have a global 'deadletter' queue which we do not want to namespace, but connect to verbatim.

This PR allows `exact: true` to be added when setting the queue_name in a consumer.


michaelklishin (changes requested):
This is a reasonable addition but the name "exact" does not explain what the option really does.

"Do not add namespace" may be longer but is more specific (and I am fine with longer names).

PR #395, opened 2023-07-04
Add delivery_info to properties for error handler

tlloydthwaites:
When handling an error, it would be useful to know the delivery_info such as the routing key. This adds delivery_info to message properties. This seemed the only way to supply the information without changing the error handler contract.

### DIFF
--- a/lib/hutch/worker.rb
+++ b/lib/hutch/worker.rb
@@ -75,7 +75,7 @@ module Hutch
     rescue => ex
       acknowledge_error(delivery_info, properties, @broker, ex)
       properties = Bunny::MessageProperties.new(properties.to_hash.merge(delivery_info: delivery_info))
-      handle_error(properties, payload, consumer, ex)
+      handle_error(properties, payload, consumer, ex, delivery_info)
     end
 
     def with_tracing(klass)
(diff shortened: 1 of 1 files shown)


michaelklishin:
Message properties and `delivery_info` are not the same thing (semantically in the protocol and in Bunny). For that reason, I don't like merging them together.

I need to take a look at what other options we may have, although very likely that'd mean modifying the error reporting interface :(

tlloydthwaites:
Yep, it's a hack... I don't see any other option other than updating the interface.

One solution might be to check `arity` on the handler method to see if it allows the extra param.

Something like:

```
handle_error(properties, payload, consumer, ex, delivery_info)
```

then:

```
backend.handle args.first(backend.method(:handle).arity)
```

This relies on delivery_info being the last parameter in the handler. We could also check using something like:

```
if backend.method(:handle).parameters.map { |e| e[1] }.include?(:delivery_info)
```

But the arity method would allow the extra param on the end without disrupting existing handlers.

tlloydthwaites:
Updated PR with arity check, working nicely here :)

michaelklishin (approved):

michaelklishin:
I still don't find the merge of the two properties to be a good idea but for error reporting and a set of existing error handlers, I guess, this may be the most pragmatic option 😭

PR #402, opened 2024-11-08
Relax activesupport dependency to support rails 8

drobny:
In addition, upgrade the GH actions checkout dependency to the latest so it runs on supported version of node

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: Test
 
-concurrency: 
+concurrency:
   group: ${{ github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
@drobny I assume it would be a good idea to produce a new patch release soon to allow for easier adoption of Hutch with new Ruby on Rails versions. Am I correct?

drobny:
> @drobny I assume it would be a good idea to produce a new patch release soon to allow for easier adoption of Hutch with new Ruby on Rails versions. Am I correct?

Yes I think that would be ideal 👍 Thanks again for your swift review and ongoing maintenance of hutch 😄

michaelklishin:
@drobny understood. You should have a new version up on RubyGems by Monday :) Have a good one and thank you for contributing!

PR #404, opened 2024-12-11
Remove upper dependency version limit on activesupport

drobny:
With the release of Rails 8.0.0.1 the dependency update I made as part of https://github.com/ruby-amqp/hutch/pull/402 is too strict and the gem gets downgraded to version 0.23.1 as part of a rails upgrade

It might be better to remove the upper ceiling on `activesupport` so this doesn't become a recurring blocker for future rails releases

### DIFF
--- a/hutch.gemspec
+++ b/hutch.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |gem|
   end
   gem.add_runtime_dependency 'carrot-top', '~> 0.0.7'
   gem.add_runtime_dependency 'multi_json', '~> 1.15'
-  gem.add_runtime_dependency 'activesupport', '>= 4.2', '<= 8'
+  gem.add_runtime_dependency 'activesupport', '>= 4.2'
 
   gem.name = 'hutch'
   gem.summary = 'Opinionated asynchronous inter-service communication using RabbitMQ'


michaelklishin:
@drobny thank you! Do we need a new urgent-ish release? I assume a new Rails major can result in a non-trivial wave of upgrades?

drobny:
> @drobny thank you! Do we need a new urgent-ish release? I assume a new Rails major can result in a non-trivial wave of upgrades?

@michaelklishin that would be amazing if you could 🙏

michaelklishin:
@drobny [done](https://rubygems.org/gems/hutch). Thank you for contributing!

drobny:
> @drobny [done](https://rubygems.org/gems/hutch). Thank you for contributing!

Thank you so much ❤️

rvall-sondermind:
@michaelklishin any chance we could get this reflected in the main changelog? Rails 8.x support is currently listed as a part of 1.4.0, not 1.3.1.

michaelklishin:
@rvall-sondermind well, it is listed in the [`1.3.x` release notes](https://github.com/ruby-amqp/hutch/blob/v1.3.x/CHANGELOG.md#131-dec-11-2024) (on that respective branch).

michaelklishin:
@rvall-sondermind I have created a GitHub release for [`1.3.1`](https://github.com/ruby-amqp/hutch/releases/tag/v1.3.1), that should be a lot more visible than the change log file on a specific long[er]-lived branch.

PR #407, opened 2025-01-07
Make `ostruct` an explicit dependency

RemoteCTO:
Addresses the following raised issue: 
 - https://github.com/ruby-amqp/hutch/issues/406

ostruct is being removed as a default gem from Ruby 3.5+, see also:
 - https://github.com/ruby/ruby/pull/10428
 - https://github.com/rails/jbuilder/issues/561
 - https://github.com/mileszs/wicked_pdf/issues/1116

### DIFF
--- a/lib/hutch/version.rb
+++ b/lib/hutch/version.rb
@@ -1,3 +1,3 @@
 module Hutch
-  VERSION = '1.4.0.pre'.freeze
+  VERSION = '1.4.1.pre'.freeze
 end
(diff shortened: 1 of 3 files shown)


lib/hutch/version.rb
```diff
@@ -1,3 +1,3 @@
 module Hutch
-  VERSION = '1.4.0.pre'.freeze
+  VERSION = '1.4.1.pre'.freeze
```
michaelklishin, line 2: Please roll this change back. This PR has nothing to do with version changes and it's up to the maintainer to decide when to change the version and why.

michaelklishin (changes requested):

michaelklishin:
Thank you!

PR #408, opened 2025-01-15
Add compatibility for ddtrace and datadog gems

Youngv:
From https://rubygems.org/gems/ddtrace
> NOTICE: This gem has been renamed to `datadog` since 2.0.0. Please use `datadog` instead of `ddtrace`. ddtrace is Datadog's tracing client for Ruby. It is used to trace requests as they flow across web servers, databases and microservices so that developers have great visiblity into bottlenecks and troublesome requests.

Add support for the new `datadog` gem.


michaelklishin:
Thank you!

PR #409, opened 2025-03-23
Cancel channel consumers before work_pool shutdown

spyroska:


spyroska:
fixes https://github.com/ruby-amqp/hutch/issues/410

michaelklishin (changes requested):
There are several things in this PR that I cannot accept.

All log messages should not be at the `info` level, or you will end up with potentially hundreds of lines that are not particularly useful.

If a consumer cannot be cancelled, retrying it in a loop, leave alone forever (without any kind of termination condition) does not make much sense.

Log messages from "a broker' do not make sense to me because the Broker is RabbitMQ, not a part of Hutch, and we should at the very least not propagate this unfortunate naming.

lib/hutch/broker.rb
```diff
@@ -382,5 +384,32 @@ def consumer_pool_size
     def consumer_pool_abort_on_exception
       @config[:consumer_pool_abort_on_exception]
     end
+
+    def cancel_all_channel_consumers
+      return unless channel&.consumers
+
+      logger.debug("Broker: Cancel channel.consumers: #{channel.consumers.inspect}")
```
michaelklishin, line 391: This should say "Connection", not "Broker".

lib/hutch/broker.rb
```diff
@@ -382,5 +384,32 @@ def consumer_pool_size
...
+
+      channel.consumers.each_value { |consumer| cancel(consumer) }
+    end
+
+    #
+    # Cancel a consumer with retry.
+    #
+    # @param [Bunny::Consumer] consumer
+    #
+    def cancel(consumer)
```
michaelklishin, line 401: This entire function is not necessary, it does not do anything useful over `Bunny::Consumer#cancel`, the retries are infinite and can prevent the process from shutting down.

Retrying consumer cancellation right before all channels and connections shut down does not make much sense, regardless of what Sneakers might do (Pika also cancels consumers on shutdown and it resulted a unique, Pika-speicific issues in certain cases).

Finally, the consumer cannot be `nil`, you are calling this method while iterating on a collection that supposedly cannot be accessed concurrently.

michaelklishin:
I have added a setting to Bunny that instructs to to explicitly cancel all consumers before shutting down in https://github.com/ruby-amqp/bunny/commit/05293634ffc0424d4dc5e0b174e8b87c75836f14.

```ruby
ch = connection.create_channel.configure do |new_ch|
  new_ch.cancel_consumers_before_closing!
end
```

It should be enough for Hutch, Kicks to enable it if they choose to do so. I don't have an opinion on whether they should, most client libraries do not do that and together with manual acknowledgements and automatic requeueing it somehow works out just fine.

But with a small number of consumers and channels, well, why not. I don't see how it would make anything different for the messages with TTL but if a project chooses to deal less or not at all with any last moment in-flight deliveries, this option makes it easier to get there.

michaelklishin:
@spyroska Hutch now depends on Bunny `2.24.0` with the above option.

You are welcome to submit a PR that would adopt it for all channels used by Hutch. Let me know if that was sufficient
for your needs.

michaelklishin:
Well, now we have to introduce the same option to March Hare, I guess. It won't be too difficult.

spyroska:
@michaelklishin Thanks for your prompt response and for trying to resolve the issue at a lower level (bunny). Let me share my thoughts on this approach. [By introducing  the Bunny option `cancel_consumers_before_closing`](https://github.com/ruby-amqp/bunny/commit/05293634ffc0424d4dc5e0b174e8b87c75836f14#diff-df2c93872027d73da6f9de309fd0d79ff4e3b24209d7c3c0483d8d92be09a326R273), which cancels consumers before closing the channel and then shutting down the pool, it is possible to do (almost) the same thing as the proposed solution in this PR. Unfortunatelly, there seems to be an (unverified) race condition in this approach: Calling `Bunny::Channel.close` will close the channel *before* draining the `Bunny::ConsumerWorkPool`. This means that some messages that are concurrently being handled and which need to use the channel to send a manual acknowledgment (at least) will fail to do so, due to the closed channel, right?. Also, in RPC scenarios, message handlers need to be able to send replies, which requires an open channel.

> Note, in this now-closed PR, the channel is available during graceful shutdown, so this race condition does not exist.

In a graceful shutdown procedure, I believe all resources (such as the channel) must be available until the work pool is drained, so that messages can use it if they need to.

Therefore, although I want to open a hutch PR that employs `cancel_consumers_before_closing`, I am afraid I cannot come up with a fix that is fundamentally different than this PR's  (a.k.a a fix that avoids manually cancelling the consumers from the hutch side). 

If you insist on fixing this in Bunny side, which indeed has the benefit of avoiding re-inventing the wheel by clients, I feel a closer look will be needed.

PR #412, opened 2025-05-20
Drop dependency on `ostruct`

Earlopain:
The only usage in library code was to hold the config and that is never exposed to users. Just use a normal struct for that.


michaelklishin (approved):

michaelklishin:
@Earlopain I'm not against using a regular struct but just out of curiosity: what motivates you to spend time removing `ostruct` from Hutch?

Earlopain:
Nobody should be using openstruct if they can help themselves. It emits a performance warning (disabled by default, you need to explicitly opt in) when used:

```sh
$ ruby -W:performance -rostruct -e "OpenStruct.new"
-e:1: warning: OpenStruct use is discouraged for performance reasons
```

Of course this is not at all a problem here. But the replacement is trivial and one less dependency is nice to me as well.

